PR #59, opened 2014-09-27
Upgrade user interface

legordian:
Start changing the user interface by moving a lot of scripts and executables into a legacy folder. Start preparing for their replacement by adding a number of functions which are exportet to python. Unfortunately, the replacement functions in python are still missing, which conerns mainly the now missing calcAmplitudes executable. However, calcAmplitudes.py is still functional, so one should be able to do everything with this version after some digging.


legordian:
Outdated and therefore closed.

PR #70, opened 2015-03-25
src: Fix a bug in pwaLikelihood's readDecayAmplitudes.

legordian:
Fix a bug in pwaLikelihood::readDecayAmplitude, where the array containing the
decay amplitudes was resized every time a new amplitude file was read in, while
slowed down the initial loading of amplitudes by some order of magnitudes.
While doing that, also finally fix the bug that the firstWave boolean was never
set to false so the consistency in terms of number of events in the amplitude
files was never properly checked as was intended.

### DIFF
--- a/src/pwaLikelihood.cc
+++ b/src/pwaLikelihood.cc
@@ -1160,9 +1160,11 @@ pwaLikelihood<complexT>::readDecayAmplitudes(const string& ampDirName,
 				  amps.push_back(amp);
 				}
 			}
-			if (firstWave)
+			if (firstWave) {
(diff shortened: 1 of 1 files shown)


suhlatwork:
Merged.

Did you test the performance impact? I cannot imagine that boost does not check the size before actually allocating memory and copying stuff around, that would just be too stupid.

legordian:
Obviously, the impact depends heavily on the number of amplitudes being loaded. I did not make a quantitative test for the 1546 waves wave set, but my guess would be that there is a factor of about 500 between before/after the fix.

Issue #25, opened 2014-08-28
[Bug] fitresult::spinDensityMatrixElem() returns non-zero value for waves with different reflectivities [sf#25]


legordian:
**Reported by bgrube on 2014-01-15 15:18 UTC**
Due to the block-diagonal form of the spin-density matrix in the reflectivity, the spin-density matrix element (interference term) for two waves with different reflectivity is zero. In the current implementation of fitresult::spinDensityMatrixElem() this is not checked and the function might return a non-zero value. This has consequences when fitResult::intensity(const char\* waveNamePattern) is called with a wave-name pattern that includes both positive and negative reflectivity (like e.g. "\*" for the total intensity). fitResult::intensity() also does not treat the flat wave as a special case.

At the moment the only solution is to parse the reflectivity value from the wave name.

suhlatwork:
Supposedly fixed with cce54d667bec5b31a94fbfb92b4a9c83bfbbedbd

Issue #31, opened 2014-08-28
[Bug] evidence calculation is wrong for certain cases [sf#31]


legordian:
**Reported by legordian on 2014-08-19 18:31 UTC**
As the determinant of the normalization matrix is wrong if there are thresholded waves, the evidence is wrongly calculated if there are waves with thresholds. The same is true for fits with both reflectivities (see also [bugs:#25]).

suhlatwork:
Supposedly fixed by 5cd064443d002802db50e8e089a439d1ae227352.

PR #75, opened 2015-04-21
Fix issues around normalization matrices

suhlatwork:
Fix issues #25 and #31.

Set the spin-density matrix elements to 0 for different reflectivities.
Set the entries in both norm matrices to 0 for different reflectivities mimicking the behaviour in the likelihood.

Properly split accepted norm matrix when calculating the evidence for fits with thresholds and ranks.

### DIFF
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -45,6 +45,7 @@
 #include "TRandom3.h"
 
 #include "fitResult.h"
+#include "partialWaveFitHelper.h"
 
 
 using namespace std;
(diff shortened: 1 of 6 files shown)


partialWaveFit/fitResult.cc
```diff
@@ -281,6 +348,12 @@ complex<double>
 fitResult::spinDensityMatrixElem(const unsigned int waveIndexA,
                                  const unsigned int waveIndexB) const
 {
+	// spin density matrix element is 0 if the two waves have different
+	// reflectivities
+	if (partialWaveFitHelper::getReflectivity(_waveNames[waveIndexA])
+	    != partialWaveFitHelper::getReflectivity(_waveNames[waveIndexB]))
+		return 0;
```
legordian: Could you please add { } here? It is confusing, as the if clause already spans two lines...

legordian:
Merged, thank you!

PR #77, opened 2015-04-23
Various fixes and improvements

suhlatwork:
* Fix building with 'Apple LLVM 6.1.0' (Xcode 6.3):
 * ignore warnings concerning ignored qualifiers while building the ROOT dictionaries (for ROOTv5 only)
 * test whether the compiler understands thread_local (independent from using -std=c++11) because Apple does not support thread_local yet
* remove one test and definition not used anymore
* add BAT and NLopt to 'printLibraryInfo'

* add a function to the partialWaveFitHelper to extract Eigenvalues and -vectors from covariance and Hessian matrices taking into account fixed parameters, use this across the code; enable 'eigenvectorLikelihoodSlices' to (properly; i.e. no bogus plots for the last eigenvalues) work with fixed parameters
* draw parabolas in 'eigenvectorLikelihoodSlices'

* clean the code reading and writing the configuration file of the resonance fit
  * be less paranoid about the in memory configuration changing while running, this should never happen
  * put the functionality needed to read extra information from one decay channel into an own method, not to have it implemented multiple times (e.g., for 'decaychannels' and 'extradecaychannels'), also in 'init' of the individual components it is no longer necessary to loop over the 'decaychannels' again

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,14 @@ if(DEBUG_OUTPUT)
 endif()
 
 
+# test whether thread_local is understood
+check_cxx_source_compiles("#include <iostream>
+int main() { thread_local int a = 1; std::cout << a << std::endl; }" COMPILER_PROVIDES_THREAD_LOCAL)
+if(COMPILER_PROVIDES_THREAD_LOCAL)
+       add_definitions(-DCOMPILER_PROVIDES_THREAD_LOCAL)
+endif()
+
+
 # redirect output files
 message(STATUS "")
 message(STATUS ">>> Setting up output paths.")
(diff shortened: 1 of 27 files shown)


suhlatwork:
Update with some more commits:

* change to format of the parameters of the final-state mass-dependence in the configuration file to be more like the one of components, e.g.:
 * do not define the five arrays for one component each, but one (libconfig) group per parameter with its own set of (the up to six) variables for one parameter
 * only 'val' and 'fix' need to be given
 * all other parameters ('lower', 'upper', 'step', 'error') are read when they are available
 * obviously this requires old files to be updated, it has been done for the example file in the repository

suhlatwork:
Update with some more commits:

* fix the 'compileBoostLibraries' script and require a more recent Boost version (after the change that broke this script)
* work around presumably a compiler bug in GCC 4.6.3 trigger inside Boost by ignoring a warning for one source file

legordian:
Merged, thank you very much!

PR #79, opened 2015-04-26
src: Make width of half-Cauchy prior settable.

legordian:
Make the width of the half-Cauchy prior changeable at runtime by making it a
standard member variable of the pwaLikelihood (as opposed to a const static
one) and adding a getter and setter for it. Also introduce a new command line
option in pwaNloptFit.cc which allows to set the prior width.

### DIFF
--- a/nlopt/pwaNloptFit.cc
+++ b/nlopt/pwaNloptFit.cc
@@ -92,7 +92,7 @@ usage(const string& progName,
 	     << endl
 	     << "usage:" << endl
 	     << progName
-	     << " -l # -u # -w wavelist [-d amplitude directory -R -o outfile -s seed -x [startvalue] -N -n normfile"
+	     << " -l # -u # -w wavelist [-d amplitude directory -R -o outfile -s seed -x -V [startvalue] -N -n normfile"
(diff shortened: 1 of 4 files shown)


suhlatwork:
Why not combine the "-C" and "-P" options (similar to what is done for "-x" of "pwafit":
* "-C" activates the Cauchy-Priors (with the default width)
* "-C x" activate the Cauchy-Priors with width x

legordian:
Ok, I looked into the possibility of combining the "-C" and "-P" options and the problem is that it does not work (at least on my box): optional arguments to options are a non-POSIX GNU extension which might not be supported everywhere.

What is much worse how optional arguments have to be specified: if you go the intuitive way and supply "-C 0.2", then optarg will be set to 0 by getopt, because separating with a space is not supported. If I do "-C=0.2", then optarg is set to "=0.2". I can also do "-C0.2" and then optarg will be set to "0.2". The same is the case for the "-x" option.

Obviously, that is very bad because one might set some options which are then not being set. I am in favor of removing all "::" options from all getopts in ROOTPWA.

legordian:
Ok, I removed the "::" arguments from pwaNloptFit and pwafit. I also included a very small cosmetic change in pwaNloptFit which should not change any functionality, but before starting a separate pull request, I am including it here.

PR #83, opened 2015-05-04
Make ROOTPWA compile with -Wextra and change default build type.

legordian:
Change ROOTPWA's default build type to "RELWITHDEBINFO", add the possibility to compile with -Wextra and fix the compilation with that.

### DIFF
--- a/amplitude/decayTopology.cc
+++ b/amplitude/decayTopology.cc
@@ -84,6 +84,7 @@ decayTopology::decayTopology(const productionVertexPtr&          productionVerte
 
 
 decayTopology::decayTopology(const decayTopology& topo)
+	: decayGraph()
 {
 	*this = topo;
 }
@@ -995,8 +996,8 @@ decayTopology::buildInternalData()
 
 interactionVertexPtr
 decayTopology::cloneNode(const nodeDesc& nd,
-                         const bool      cloneInParticles,
-                         const bool      cloneOutParticles)
+                         const bool,
+                         const bool)
 {
 	const interactionVertexPtr  vert        = vertex(nd);  // this must not be a reference
 	const bool                  isProdVert  = isProductionVertex(vert);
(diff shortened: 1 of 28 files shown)


bgrube:
Great! Looks good.
Thanks!

amplitude/decayTopology.cc
```diff
@@ -995,8 +996,8 @@ decayTopology::buildInternalData()
 
 interactionVertexPtr
 decayTopology::cloneNode(const nodeDesc& nd,
-                         const bool      cloneInParticles,
-                         const bool      cloneOutParticles)
+                         const bool,
+                         const bool)
 {
```
suhlatwork: should probably not be removed here, but the parameters should be added to the call to cloneNode in line 1004 (old revision).

amplitude/test/testPeakShapeInstableIsobar.cc
```diff
@@ -272,7 +272,7 @@ class threeBodyDynAmpMc : public threeBodyDynAmpInt {
 
 public:
 
-	threeBodyDynAmpMc(const dynamicAmplitude& isobarAmp,
+	threeBodyDynAmpMc(const dynamicAmplitude&,
 	                  const double            mass         = 0,
 	                  const double            width        = 0,
 	                  const unsigned int      L            = 0,
```
suhlatwork: parameter should not be removed, but added to the call of the constructor of the base class (I wonder why this ever compiled, probably because it was not used anywhere and optimised away)

amplitude/waveDescription.cc
```diff
@@ -1029,7 +1029,7 @@ waveDescription::writeKeyFile(FILE&                      outStream,
 bool
 waveDescription::writeKeyFile(FILE&                  outStream,
                               const isobarAmplitude& amplitude,
-                              const bool             writeProdVert)
+                              const bool                       )
```
suhlatwork: parameter should not be removed but probably be added to the call to setKeysFromAmplitude in line 1036

suhlatwork:
Even though this has already been merged, this pull request seems to be incomplete:
* following the changes to evtTreeHelper.(cc,h) also convertTreeToEvt.C needs to be modified
* in the _v2 branch we still have the wrong usage of the SYSTEM statement in Cmake's include_directories command. In particular this pull request misses a number of warnings originating from the header-only part of the utilities directory

PR #85, opened 2015-05-06
Yaml

suhlatwork:
Replace usage of libconfig in the mass-dependent fit by yaml.

The used library is yaml-cpp, it is required to build ROOTPWA.


legordian:
These commits have been merged after a rebase, so github does not detect the merge. Closed.

PR #87, opened 2015-05-11
src: Add getter for fit parameter error.

legordian:
Add a getter for the fit parameter error by name to the fit result.


partialWaveFit/fitResult.h
```diff
@@ -137,6 +137,7 @@ namespace rpwa {
 
 		double      fitParameter   (const std::string& parName  ) const;  ///< returns value of fit parameter with name
 		/// returns covariance of fit parameters at index A and B
+		double      fitParameterErr(const std::string& parName  ) const;  ///< returns error of fit parameter with name
```
suhlatwork: position of comment and this line should be changed

PR #92, opened 2015-05-20
Various fixes and constant background for resonance fit

suhlatwork:
* add a constant background parametrisation to the resonance fit
* correct some types of variables in relativisticAmpCorrections and resonanceFit
* cosmetics: better names of minimizer parameter names for FSMD, ...


suhlatwork:
* add a (purely virtual) method to get the number of free parameters in a fit to the minimizer interface, up to now such a method could have been named differently for different interfaces
* implement this method for the ROOT interface

suhlatwork:
* the 'freeparameters' option in the configuration file must not be an empty sequence, it has to either not exist (then default values are used) or it has to contain useful stuff

PR #86, opened 2015-05-11
Usr iface

odrotleff:

PR #93, opened 2015-06-07
Usr iface II

suhlatwork:
This pull request is supposed to replace #86. It is based on the changes of this pull request including commit 26e34ae1428d00dc559569474e8f2fa72143fb4c.

Compared to this the following changes have been done:
* add an accessor to the `TTree` pointer of an `amplitudeMetadata` object (similar to `eventMetadata`)
* use (pointers to) `amplitudeMetadata` objects in a couple of places instead of pointers to `TTree` and wave names:
 * `ampIntegralMatrix::integrate`
 * `pwaLikelihood::addAmplitude`
* move the several `legacyScripts` directories to the top level directory `legacy`, the idea would be to have stuff that still compiles directly in this directory, and stuff that does no longer compile in the same structure as in the current main directory, but below `legacy/`
 * add back C++ executables `calcAmplitudes`, `calcIntegrals`, and the two `pwafit`s using the new file formats and facilities (except the file manager (obviously)), this should ease the migration a bit in case where some parts in the file manager are still missing
* update `getReflectivity` to work with old and new key file naming convention, as this is still required in the `fitResult`
* `pwaLikelihood::init` now requires a vector of tuples of wave name, wave description and threshold (instead of two separate vectors), the wave name has been added, it is required if the name of the on-file object does not correspond to the name constructed from the decay topology (this seems to be the simplest way to keep the possibility of having the same waves with different isobar parametrization (though I did not try this till the end))
* once again `_decayAmps` was resized for each new amplitude
* entirely (I hope so) remove some stuff no longer consistent with the new user interface:
 * binary amplitude files
 * old key file name convention
* similarly to `calcAmplitudes` allow to limit the bins for `calcIntegrals`
* check that a valid event type was specified for `calcIntegrals`
* do not stop if a integral file already exists, but simply skip (as done for the amplitudes)

Quite a number of smaller bugfixes and cosmetics:
* added include guards
* increase similarity between ROOT and NLopt fitter
 * add option to not calculate and store Hesse matrix for ROOT
 * add Cauchy priors for ROOT
 * check analytic Hesse matrix without storing it for NLopt
* remove further usages strings for `amplitude` and `integral` (names for ROOT objects) and replace by the static variables of the according meta data class
* some stuff (mostly cosmetics) lost in the rebases
* build system:
 * treat some more libraries/subdirectories in the new standard way
 * sort the various include paths
* increase `ClassDef` version of `eventMetadata`

I basically reshuffled all the commits, most commits should compile (again), 1504fca4be3e53d94f8060847fc1cd7c73029848 is the last commit largely equivalent to the current master.

I tested this branch on a single bin of my data obtaining the same result as with the current master, I do not know how to test this more properly (in any case some of the test scripts were broken?). This is a comparison of the memory usage:

 | master (binary amplitudes) | master (ROOT amplitudes) | branch C++ | branch py
----|---|---|---|---
calcAmplitudes | 165 MB | 415 MB | 415 MB | 420 MB
calcIntegrals | 6 MB | 1,45 GB | 1,75 GB | 1,75 GB
pwafit (ROOT) | 120 MB* | 530 MB | 376 MB | 396 MB
\* during writing up to 350 MB

So `calcIntegrals` still needs to be improved, but that was already the case with ROOT amplitudes previously. A couple of issues raised in #86 are still open, but they can also be attacked later.


suhlatwork:
Small update:
* move default arguments from the implementation to the declaration for pwaFit high-level interface
* do not have default arguments for the implementation of the function called by Boost for `calcAmplitude`, the default arguments are defined when adding this function to the Boost Python interface.
* do not have wrapper functions for the `pwaFit` classes, with the new interface Boost can figure this out automatically
* improve the extraction of a Boost tuple from a Boost Python tuple
* make both `CovarianceMatrix` methods of `pwaLikelihood` callable, previously always the one taking a TMatrixT was chosen.

Issue #91, opened 2015-05-18
Numerical instabilities in "isobarAmplitude::gjTransform"


suhlatwork:
"isobarAmplitude::gjTransform" never checks whether the beam and X Lorentz-vectors actually span a production plane or are collinear.

A proper solution probably has to handle two cases:
* beam and X are collinear and the three-vector component is not equal to zero
* beam and X are collinear with a zero three-vector component
Case one can probably not be handled without spoiling theta_GJ and phi_TY, the second case can probably be handled by simply not performing any transformation (as X already appears to be in its centre-of-mass system) and relying on the user to have taken care that those two angles still are correct. However, the check for equal to zero needs to include some numerical margin, as typically X is calculated from its daughters.

This is actually a real problem, as the "phaseSpaceIntegral" class uses a non-interaction vertex which can only provide equal beam and X Lorentz-vectors to calculate the amplitudes.

bgrube:
Oops, but the how does phaseSpaceIntegral handle the t' dependence?

suhlatwork:
For the pure phaseSpaceIntegrals (without taking acceptance into account) there should be no t' dependence?

bgrube:
Yes, I was confusing it with the phase-space integral we use for the likelihood.

PR #95, opened 2015-06-19
Fix phaseSpaceIntegral and small stuff

suhlatwork:
Fix the values calculated by the `phaseSpaceIntegral` class. The volume of the integration was taken into account twice, once in the normalization of the `S_U_CHUNG` weight, and once explicitly in the calculation this class. Now only use the one from the normalization.

Also issue #91 is partially fixed. If the two Lorentz-vectors of beam and X are equal and X is at rest (a small variation is allowed), the boost matrix, which is basically the unit matrix except for the small allowed variation, is returned. The boost is also performed if X is not at rest, however, then a warning is printed.

The effects of those two corrections are depicted here:
plots

Also fix a couple of smaller stuff:
* the copy and assignment constructors of the `waveDescription` should only parse the key file content, if they were successfully parse before, otherwise it is not possible to copy an empty `waveDescription` object without an error message.
* fix to compile with `-Wextra`
* fix normalization of amplitudes in view of zero entries in the normalization integral, zero entries might occur when using the flatRange mass-dependence, however, then integral and amplitude should be zero. So in case of zero amplitude and integral do nothing (the amplitude remains zero (but does no longer become NaN spoiling everything)), in case of zero integral and non-zero amplitude print an error and exit because this should neverhappen, and in the case of non-zero integral do the normalization. (Note that this was rather poorly fixed in case of the binary amplitudes already before.)

### DIFF
--- a/decayAmplitude/isobarAmplitude.cc
+++ b/decayAmplitude/isobarAmplitude.cc
@@ -136,6 +136,25 @@ TLorentzRotation
 isobarAmplitude::gjTransform(const TLorentzVector& beamLv,  // beam Lorentz-vector
                              const TLorentzVector& XLv)     // X  Lorentz-vector
 {
+	if (beamLv == XLv) {
+		// Beam and X Lorentz vector are equal if the amplitude is to
+		// be calculated starting with a non-interaction vertex as it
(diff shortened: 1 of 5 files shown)


legordian:
Thank you very much for these fixes! Merged.

PR #96, opened 2015-06-22
Some fixes and stuff for resonance fit

suhlatwork:
Fixes:
* fix partial-wave fitters in `legacy`, on Mac a missing colon after a parameter definition for `getopt` does apparently not matter, for Linux it definitively does
* explicitly load the `std::comples` dictionaries in `calcIntegrals`

Cosmetics and consistency:
* remove the option to write ASCII integral matrices with `calcIntegrals` from `legacy`
* replace `norm.int` by `norm.root` in a couple of places (help texts and default values) (also at places where the ROOT version had been required before)
* remove all include statements from the header files in `pyInterface` only declaring a single `export...` function, those includes are not required to parse and understand the header files, but are only required in the implementation
* rename `utils.*` to `reportingUtilsEnvironment.*`
* add `breakupMomentum` and `barrierFactorSquared` to the `pyInterface` (those functions have been chosen because I needed them)
* update wavelists and scripts generating the keyfiles for pi- pi- pi+, pi- pi0 pi0, and pi- eta eta to the new naming scheme for waves
* fix the `pyRootPwa`  test executed by `make test`, the test still expected a wave name according to the old naming scheme, the expectation has been updated to the new naming scheme

Resonance fit:
* remove the parameter to shift the mass used to calculate the break-up momentum in `exponentialBackground`
* add option to set the exponent of the break-up momentum in the background parameterisations
* add option to add a barrier factor using an angular momentum from the options file to the break-up momentum `(q*F^2(l))^exponent`
* add new background parametrisations using the phase-space integral instead of the break-up momentum (and the barrier factors)
* normalise `q*F^2` and the phase-space integral at the highest mass-bin, this should make the parameters for the usual backgrounds and the backgrounds using the phase-space integrals somewhat comparable, but this will render previous fit results to be no longer correct


legordian:
Very nice, thank you! Merged!

PR #97, opened 2015-06-26
Various things

suhlatwork:
* continue to remove the old binary amplitude file format and change the ROOT format over to using metadata
 * change a couple of comments
 * remove the option to read binary amplitude files and ASCII integral matrices from `weightEvents`, use `amplitudeMetadata` to get the tree containing the amplitudes

* new wave naming scheme in
 * `weightEvents`
 * `genPseudoData.py` (exporting `partialWaveFitHelper::getReflectivity` for this)

* move the initialization of `stdoutisatty` and `stderrisatty` so that it is done while `import pyRootPwa` so that not every script needs to set this

* phase-space integral used in the mass dependent fit now is the real phase-space integral without the factor `mass` that Dima includes

* correct various error messages and warnings


legordian:
Thank you very much, merged!

PR #99, opened 2015-07-13
partialWaveFit: Fix wave name escaping.

legordian:
Fix the treatment of the TString with the wave name which is fed to the
TPRegexp so that the new wave name convention works as well.

### DIFF
--- a/partialWaveFit/fitResult.h
+++ b/partialWaveFit/fitResult.h
@@ -494,8 +494,12 @@ namespace rpwa {
 	{
 		// escape special characters:
 		TString Pattern(waveNamePattern);
-		Pattern.ReplaceAll("+","\\+");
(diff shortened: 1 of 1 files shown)


suhlatwork:
I would prefer not merging this:
* escaping the special characters used for regular expressions at one point starts to hinder the usage of regular expressions, for that reason I would prefer not to do any escaping here (and even remove the current one), users of this functions can always pass an already escaped wave/amplitude name
* a function to escape the characters special for regular expressions already exists (`escapeRegExpSpecialChar`) but needs slight improvements ('^' is handled twice, double escapes '\\' are not collapsed to one '\') but should be used instead of listing all characters here again

legordian:
You are right, closed without merging. See also Issue #102 and Issue #101.

PR #100, opened 2015-07-29
pyInterface: improve interface of userInterface/calcAmplitudes.py

odrotleff:
improve interface of calcAmplitudes script to accept a keyfile index for easy
parallelization.


suhlatwork:
Could you please check that '-k' and '-w' are not specified at the same time?

suhlatwork:
Merged with one additional change: the check for an empty `waveList` is performed after the `keyfileIndex` parameter was checked, and not before.

PR #104, opened 2015-08-11
add tprime function from kinematic variables

fkrinner:
function added to calculate tprime (t') from kinematic variables as
TLorentzVectors. Implement this as method for the decayTopology as well
as in the python bindings to be called from python

### DIFF
--- a/pyInterface/utilities/physUtils_py.cc
+++ b/pyInterface/utilities/physUtils_py.cc
@@ -1,14 +1,29 @@
 #include "physUtils_py.h"
 
 #include <boost/python.hpp>
-
+#include <TLorentzVector.h>
 #include "physUtils.hpp"
-
+#include "rootConverters_py.h"
 namespace bp = boost::python;
 
 
+double tPrime_py(PyObject* beam, PyObject* target, PyObject* out){
+        TLorentzVector* lvBeam   = rpwa::py::convertFromPy<TLorentzVector*>(beam  );
+        TLorentzVector* lvTarget = rpwa::py::convertFromPy<TLorentzVector*>(target);
+        TLorentzVector* lvOut    = rpwa::py::convertFromPy<TLorentzVector*>(out   );
+
+        return rpwa::tPrime(*lvBeam,*lvTarget,*lvOut);
+};
+
 void rpwa::py::exportPhysUtils() {
 
+        bp::def(
+	        "tPrime"
+	        , &tPrime_py
+	        , (bp::arg("pBeam"),
+	           bp::arg("pTarget"),
+	           bp::arg("pOut"))
+	);
 	bp::def(
 		"breakupMomentumSquared"
 		, &rpwa::breakupMomentumSquared
--- a/utilities/physUtils.hpp
+++ b/utilities/physUtils.hpp
@@ -34,6 +34,7 @@
 #ifndef PHYSUTILS_H
 #define PHYSUTILS_H
 
+#include <TLorentzVector.h>
 
 #include "mathUtils.hpp"
 #include "reportingUtils.hpp"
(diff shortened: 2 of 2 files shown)


decayAmplitude/diffractiveDissVertex.cc
```diff
@@ -42,9 +42,11 @@
 #include "TObjString.h"
 #include "TVector3.h"
 
+
```
legordian: Could you please remove this newline?

decayAmplitude/productionVertex.h
```diff
@@ -76,6 +76,9 @@ namespace rpwa {
 
 		virtual std::string name() const { return "productionVertex"; }  ///< returns label used in graph visualization, reporting, and key file
 
+		virtual double getTprime() const {printWarn<<"getTprime not overwritten"<<std::endl;return 0.;};
+
+
```
legordian: Why don't you make this function abstract? Also, there is one newline too much.

pyInterface/utilities/tPrime_py.h
```diff
@@ -0,0 +1,13 @@
+#ifndef TPRIME_PY
```
legordian: This should go in pyInterface/utilities/physUtils_py.{h,cc}

pyInterface/rootPwaPy.cc
```diff
@@ -101,6 +102,7 @@ BOOST_PYTHON_MODULE(libRootPwaPy){
 	rpwa::py::exportCalcAmplitude();
 	rpwa::py::exportPwaLikelihood();
 	rpwa::py::exportPwaFit();
+
 #ifdef USE_NLOPT
```
legordian: Could you please remove these newlines, this file does not have to be changed at all.

decayAmplitude/diffractiveDissVertex.cc
```diff
@@ -336,3 +338,10 @@ diffractiveDissVertex::printPointers(ostream& out) const
 	    << "recoil particle = " << recoil()    << endl;
 	return out;
 }
+
+double diffractiveDissVertex::getTprime() const{
+	const TLorentzVector lvBeam = TLorentzVector(beam()->lzVec());
+	const TLorentzVector lvTarget = TLorentzVector(target()->lzVec()); 
+	const TLorentzVector lvOut = TLorentzVector(XParticle()->lzVec());
+	return tPrime(lvBeam,lvTarget,lvOut);
```
legordian: Could you please add spaces after the commas?

decayAmplitude/diffractiveDissVertex.cc
```diff
@@ -336,3 +338,10 @@ diffractiveDissVertex::printPointers(ostream& out) const
 	    << "recoil particle = " << recoil()    << endl;
 	return out;
 }
+
+double diffractiveDissVertex::getTprime() const{
+	const TLorentzVector lvBeam = TLorentzVector(beam()->lzVec());
```
legordian: Why are you using the copy constructors here? Also, why are you not using references?

highLevelInterface/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ endif()
 set(SOURCES
 	calcAmplitude.cc
 	pwaFit.cc
+	pwaNloptFit.cc
```
suhlatwork: This addition should be removed (it is responsible for the Travis error).

`pwaNloptFit.cc` should only be added if `USE_NLOPT` is `true`, and that is already taken care of just below.

decayAmplitude/diffractiveDissVertex.cc
fkrinner: The t' function boosts the 4-vectors, but the class method should be const, thus the vectors have to be copied. References will be added

legordian: But the t' function does not have references as argument, so it does not change 4-vectors, isn't it?

decayAmplitude/productionVertex.h
fkrinner: Since productionVertex could be much more than diffractiveDissociation, thinking e.g. of Deck-like effects, an abstract function would i.m.o. not be the optimal choice, since t' is not very meaningful for these kind of processes, and thus a corresponding errorMessage would have to be implemented everytime.
I would change the message though:
printWarn<<"getTprime not defined for current process "<<name()<<std::endl;

legordian:
Great, thank you for implementing our comments!

PR #106, opened 2015-08-19
Fix t'

suhlatwork:
Fix some stuff around the calculation of t':
* wrong direction of boost into (actually even more away from) the target-reset-frame
* replaced a hard-coded proton mass
In the end decided to use a simpler approach.


legordian:
Merged, thanks!

PR #107, opened 2015-08-24
pyInterface: short term fix for genPseudoData

odrotleff:
Fix genPseudoData in the short term. Not yet a nice implementation of the file
manager since there is currently work being done on that front. Just a quick
and dirty fix so the script works.


suhlatwork:
It appears to me, that the `getFromTDirectory` stuff is only working as long as there is only one type of objects that can be retrieved from a `TDirectory` with this method. Now that this function is making its way from stuff purely used for testing into "normal" operation, this needs to be reconsidered.

suhlatwork:
Merged after also correcting the function call in `integralBenchmark.py`. Thank you!

PR #108, opened 2015-08-25
improve memory structure for pwaLikelihood.cc

fkrinner:
improved the memory structure of the pwaLikelihood to account for
different numbers of positive and negative reflectivity waves


suhlatwork:
Do we really need the `copy*ParArray` methods with `decayAmpsArrayType` arguments? I would prefer to remove/not add those. I would also prefer to use `prodAmpsArrayType` to be a bit more consistent with the rest of the code where production amplitude stuff is always abbreviated like this.
Also please use a proper name and e-mail address.

PR #110, opened 2015-08-31
pyInterface: make genPseudoData generate eventsType "real"

odrotleff:
make genPseudoData generate events of the type "real" instead of "generated".


legordian:
Merged, thank you!

PR #111, opened 2015-08-31
Pull request for steplike isobars.

fkrinner:
steplike isobars implemented, which can be created from a single keyFile

### DIFF
--- a/decayAmplitude/massDependence.cc
+++ b/decayAmplitude/massDependence.cc
@@ -86,6 +86,17 @@ flatRangeMassDependence::amp(const isobarDecayVertex& v)
 	return amp;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+complex<double>
+steplikeMassDependence::amp(const isobarDecayVertex& v)
+{
+	const particlePtr& parent = v.parent();
+	double mIsob = parent->lzVec().M();
+	if (mIsob>_mMin and mIsob<_mMax){
+		return complex<double>(1.,0.);
+	};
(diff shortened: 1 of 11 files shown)


legordian:
lorentzFactor change: b1c3876ab9ed4e0a570f20e5a195a166ed8f7abd

suhlatwork:
A couple of things are not working with the latest state of this request:
* different `waveDescriptions` expanded from the same keyfile get the same name which is a bad thing for amplitude file names, ...
* the changes to the file manager are not in agreement with the changes done on the C++ side

I already did some changes to the code I agree with, mostly those have been cosmetic (white-spaces and other coding-style issues) except for:
* your step-like mass dependence would always return amplitude 0 for a mass directly on the bin border, now in such cases the amplitude would be 1 in the higher bin 
* the `while (true)` loop during the expansion has been replaced with recursion (which I think makes the code a bit more readable)
* `get` and `mapMassDependenceType` have been merged into one again

Please carefully check and continue working based on e544f47da9418c195e9e1c17f9017d32d47ff847.

I am still not happy with:
* the `flatRangeMassDependence` should be removed now that this better way becomes available
* the name `steplikeMassDependence` of the `massDependence` class (a step only goes either up or down), I would prefer `rectangularMassDependence` or re-using `flatRange...`
* the name of the mass-dependence given in the keyfile (now `freed-isobar`/`freedIsobar`, but strictly speaking a single bin is not a freed-isobar), as for the other mass-dependencies we should use the same name here as for the class

Let us continue working on this once you are back.

PR #112, opened 2015-09-03
docs: update and extend README

bgrube:
update instructions for Boost and yaml-cpp; remove references to
SourceForge; update instructions for ROOTPWA tags and branches

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ CMake is a cross-platform, open-source build system available from <http://cmake
 *   <http://rachid.koucha.free.fr/tech_corner/cmake_manual.html>
 *   <http://mash-project.eu/wiki/index.php/CMake%3a_Getting_Started>
 
-The minimum required CMake version is 2.8.8. In case your system offers only outdated packages (check CMake version by running `cmake --version`), you can quite easily compile CMake yourself...
+The minimum required CMake version is 3.0.0. In case your system offers only outdated packages (check CMake version by running `cmake --version`), you can quite easily compile CMake yourself...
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -120,18 +120,20 @@ We use the _libconfig_ config file parser written by Mark A. Lindner available f
...
+As YAML parser we use _yaml-cpp_ written by Jesse Beder available from <https://github.com/jbeder/yaml-cpp>. Version 0.5 or higher is required. Unfortunately there is no way to check the version in the current release, so please do this manually. The easiest way to install yaml-cpp from source (assuming you have no administrator privileges) is: 
 
-    > cd ${YAML_CPP_BUILD}
-    > cmake ${YAML_CPP_SOURCE} -DCMAKE_INSTALL_PREFIX=${YAML_CPP_INSTALL} -DBUILD_SHARED_LIBS=ON
-    > make && make install
+	> git clone https://github.com/jbeder/yaml-cpp.git
+    > cd yaml-cpp
+    > mkdir build; cd build
+    > CXXFLAGS="-fPIC" cmake .. -DBUILD_SHARED_LIBS=ON
+    > make
```
suhlatwork: Is setting `CXXFLAGS` really required? For me setting `BUILD_SHARED_LIBS` was always enough, but one needed to set this when running `cmake` for the first time, because it would never update existing cmake output.

One might also mention `-DBoost_NO_SYSTEM_PATHS=ON` which seems to be required on a couple of systems if Boost is installed in `/usr` by the system and somewhere else by the user.

legordian:
Merged, thank you!

PR #113, opened 2015-09-03
libconfig: fix compilation error with V1.5.0

bgrube:
the signature setting[std::string] was removed; fix code to use
setting[const char*] instead; tested to work with V1.4.9


legordian:
Merged, thank you!

PR #115, opened 2015-09-03
buildSystem: get version for YamlCpp

suhlatwork:
Try to get a version for YamlCpp from the pkgconfig file installed by
the library.


legordian:
Merged, thanks!

PR #116, opened 2015-09-09
Fix Nlopt fit

suhlatwork:
* Import some rather important fix to the Nlopt partial-wave fit by Oliver
* some cosmetics (whitespaces) in waveDescription.cc`


legordian:
Merged, thank you!

PR #119, opened 2015-09-14
Fix a couple of things concerning mass dependencies and keyfiles

suhlatwork:
* use same mass dependence names when writing and reading keyfile
* allow to store `flatMassDependence` for an isobar decay
* Python default value in `waveDescription::writeKeyFile` is same as in C++

### DIFF
--- a/decayAmplitude/waveDescription.cc
+++ b/decayAmplitude/waveDescription.cc
@@ -867,13 +867,14 @@ waveDescription::setXQuantumNumbersKeys(Setting&        XQnKey,
 
 bool
 waveDescription::setMassDependence(Setting&              isobarDecayKey,
-                                   const massDependence& massDep)
+                                   const massDependence& massDep,
+                                   const bool            XDecay)
 {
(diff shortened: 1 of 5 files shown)


decayAmplitude/waveDescription.cc
```diff
@@ -658,12 +658,12 @@ massDependencePtr
...
 	massDependencePtr massDep;
-	if (   (massDepType == "BreitWigner")
+	if (   (massDepType == "relativisticBreitWigner")
 	    or (massDepType == ""))  // default mass dependence
 		massDep = createRelativisticBreitWigner();
 	else if (massDepType == "constWidthBreitWigner")
 		massDep = createConstWidthBreitWigner();
-	else if (massDepType == "flat")
+	else if (massDepType == "flatMassDependence")
 		massDep = createFlatMassDependence();
```
legordian: If we use "flatRange", would not the consistent choice be "flat" here?

legordian:
Merged, thank you!

PR #120, opened 2015-09-15
partialWaveFit: fix second derivative of Cauchy prior.

suhlatwork:
Fix the second derivative of the Cauchy prior, this also fixes negative
eigenvalues in the Hesse matrix that appeared when using the Cauchy
prior.


legordian:
Merged, thank you!

PR #121, opened 2015-09-16
partialWaveFit: Clean up wave name escaping.

legordian:
Clean up the business with the pattern matching for wave names in the fit
result by removing some ancient lines, making the escaping functions available
in python and adjusting plotGui accordingly. This probably breaks some of the
root scripts using the fit result, but all the compiled code seems to continue
to work.

### DIFF
--- a/partialWaveFit/fitResult.h
+++ b/partialWaveFit/fitResult.h
@@ -62,30 +62,36 @@
 namespace rpwa {
 
 	inline
-	TString
+	std::string
 	escapeRegExpSpecialChar(const std::string& s)  ///< escapes all special characters used in regular expressions
 	{
 		TString escapedS(s);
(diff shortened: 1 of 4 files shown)


partialWaveFit/fitResult.h
```diff
@@ -71,13 +71,13 @@ namespace rpwa {
 			TString escapedChar = TString("\\").Append(specialChars[i]);
 			escapedS.ReplaceAll(specialChars[i], escapedChar);
```
suhlatwork: * '^' are still escaped twice
* you should remove double escapes
```
    const TString escapedTwiceChar = TString("\\").Append(escapedChar);
    escapedS.ReplaceAll(escapedTwiceChar, escapedChar);
```

legordian:
Thanks for the comments. Updated.

PR #122, opened 2015-09-17
pyInterface: Add some utility functions.

legordian:
Add two functions to the utilities which should make it easier to handle large
numbers of fit results in TTrees.

### DIFF
--- a/pyInterface/package/utils/_fitTreeUtils.py
+++ b/pyInterface/package/utils/_fitTreeUtils.py
@@ -1,0 +1,24 @@
+
+
+import pyRootPwa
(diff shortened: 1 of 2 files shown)


pyInterface/package/utils/_fitTreeUtils.py
```diff
@@ -0,0 +1,24 @@
+
+
+import pyRootPwa
```
suhlatwork: Isn't this a circular dependency? How does Python resolve those?

legordian: As these import are all over the package directory, it seems python is able to handle it...

PR #123, opened 2015-09-17
Executables in storageFormats

suhlatwork:
* make the Python version of `printMetadata` capable of the same things as the C++ version
* remove the C++ executables `printMetadata` and `convertEventFile`


legordian:
Nice, thank you. Merged!

PR #125, opened 2015-09-29
partialWaveFit: Remove TString from interface.

legordian:
Change all the TStrings appearing in the interface of fitResult to std::string
and adapt all affected places accordingly.

### DIFF
--- a/generators/weightEvents.cc
+++ b/generators/weightEvents.cc
@@ -338,8 +338,8 @@ main(int    argc,
 			}
 
 			for (unsigned int iProdAmp = 0; iProdAmp < result->nmbProdAmps(); ++iProdAmp) {
-				const string prodAmpName = result->prodAmpName(iProdAmp).Data();
-				const string waveName    = result->waveNameForProdAmp(iProdAmp).Data();
+				const string prodAmpName = result->prodAmpName(iProdAmp);
+				const string waveName    = result->waveNameForProdAmp(iProdAmp);
 				const int iWave          = result->waveIndex(waveName);
 
 				// extract rank and reflectivity from wave name
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -189,7 +189,7 @@ fitResult::evidenceComponents() const
 	set<unsigned int> thrWaveIndices;
 	for (set<unsigned int>::const_iterator it = thrProdAmpIndices.begin();
 	     it != thrProdAmpIndices.end(); ++it) {
-		const int index = waveIndex(waveNameForProdAmp(*it).Data());
+		const int index = waveIndex(waveNameForProdAmp(*it));
 		// printDebug << " wave[" << index << "] = '" << waveNameForProdAmp(*it).Data()
 		//            << "' has zero production amp[" << *it << "] = '"
 		//            << _prodAmpNames[*it] << "'" << endl;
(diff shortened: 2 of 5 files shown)


partialWaveFit/fitResult.cc
```diff
@@ -543,8 +543,8 @@ fitResult::normIntegralForProdAmp(const unsigned int prodAmpIndexA,
                                   const unsigned int prodAmpIndexB) const
 {
 	// treat special case of flat wave which has no normalization integral
-	const bool flatWaveA = prodAmpName(prodAmpIndexA).Contains("flat");
-	const bool flatWaveB = prodAmpName(prodAmpIndexB).Contains("flat");
+	const bool flatWaveA = TString(prodAmpName(prodAmpIndexA)).Contains("flat");
+	const bool flatWaveB = TString(prodAmpName(prodAmpIndexB)).Contains("flat");
```
suhlatwork: this should be replaced by
```
const bool flatWaveB = (prodAmpName(prodAmpIndexB).find("flat") != string::npos);
```

partialWaveFit/fitResult.cc
```diff
@@ -820,7 +820,7 @@ fitResult::fill
 		// look for index of first occurence
 		unsigned int j;
 		for (j = 0; j < _prodAmpNames.size(); ++j)
-			if (prodAmpName(j).Contains(waveName))
+			if (TString(prodAmpName(j)).Contains(waveName))
```
suhlatwork: similarly here
```
if (prodAmpName(j).find(waveName) != string::npos)
```

generators/weightEvents.cc
```diff
@@ -338,8 +338,8 @@ main(int    argc,
 			}
 
 			for (unsigned int iProdAmp = 0; iProdAmp < result->nmbProdAmps(); ++iProdAmp) {
-				const string prodAmpName = result->prodAmpName(iProdAmp).Data();
-				const string waveName    = result->waveNameForProdAmp(iProdAmp).Data();
+				const string prodAmpName = result->prodAmpName(iProdAmp);
+				const string waveName    = result->waveNameForProdAmp(iProdAmp);
```
suhlatwork: is it possible to use references here?

resonanceFit/massDepFit.cc
```diff
@@ -1606,7 +1606,7 @@ rpwa::massDepFit::massDepFit::readFitResultMatrices(TTree* tree,
 		// amplitudes of the fit result
 		std::vector<unsigned int> prodAmpIndicesForCov(_nrWaves);
 		for(unsigned int idxProdAmp=0; idxProdAmp < fit->nmbProdAmps(); ++idxProdAmp) {
-			const std::string waveName = fit->waveNameForProdAmp(idxProdAmp).Data();
+			const std::string waveName = fit->waveNameForProdAmp(idxProdAmp);
```
suhlatwork: is it possible to use references here?

partialWaveFit/fitResult.h
```diff
@@ -583,7 +583,7 @@ namespace rpwa {
 			         << "cannot deduce corresponding wave name." << std::endl;
 			return "";
 		}
-		return prodAmpName(3, prodAmpName.Length() - 3);
+		return prodAmpName(3, prodAmpName.Length() - 3).Data();
 	}
```
suhlatwork: this function `waveNameForProdAmp` should probably entirely be rewritten taking into account multi-digit ranks.

suhlatwork: this would also remove the special case for `V_flat`

```
std::string
fitResult::waveNameForProdAmp(const unsigned int prodAmpIndex) const
{
	const std::string& prodAmpName = _prodAmpNames[prodAmpIndex];
	if (prodAmpName.length() == 0 or prodAmpName[0] != 'V' or prodAmpName.find('_') == std::string::npos) {
		printErr << "production amplitude name '" << prodAmpName << "' does not follow the naming convention. "
		         << "cannot deduce corresponding wave name." << std::endl;
		return "";
	}
	return prodAmpName.substr(prodAmpName.find('_')+1);
```

legordian: Implemented, thank you very much!

PR #127, opened 2015-10-05
Fix stuff in the buildsystem

suhlatwork:
* `.so` is a valid file extension for shared libraries on Mac (at least when using ROOT)
* some cleaning of `FindROOT`


legordian:
Merged, thank you!

PR #128, opened 2015-10-08
Calculate covariance matrix for fit result in Python

suhlatwork:
* add a Python script to calculate the covariance matrix for an existing fit result, and removed the C++ equivalent (which was no longer compilable since switching to the Python user interface)
* allow to import additional variables while importing "old" event files
* work with bins that have zero real events, but have MC (generated and accepted)
* compile again with libconfig 1.5


suhlatwork:
Moved 'calcCovMatrix...' to more appropriate place and removed the zero event stuff (I need to think about this some more).

legordian:
Merged, thank you!

PR #129, opened 2015-10-13
Some more Python conversions

suhlatwork:
* add conversions of the old `eigenvectorLikelihoodSlices` and `likelihoodPointCalculator` executables that were disabled when switching to the new user interface
* move a couple of functions that are/might be used in multiple places into the `pyRootPwa.utils` namespace:
 * the function to read a `wavelist` now is called `getWaveDescThresFromWaveList` used in `calcAmplitudes` (to limit the number of waves to calculate the amplitudes for) and the two `pwafit`s (replacing two similar functions by one)
 * the function to get a wavelist from a fit result (`getWaveDescThresFromFitResult`) was moved from `calcCovMatrixForFitResult` because it is also used in the two new scripts
 * the function to get a fit result from a file (`getFitResultFromFile`) has been moved from `calcCovMatrixForFitResult` because it is also used in the two new scripts
 * the function to get the best fit result from a file (`getBestFitResultFromFile`) has been moved from `genPseudoData` because it might become useful in more places (this function simply returns the result with the best likelihood)
 * a new function `getBestFitResultsFromFile` has been added which returns a dictionary from `massBinCenters` to the best result in this mass bin
* move the code to calculate the Eigenvalues of the Hessian/covariance matrix from the `partialWaveFitHelpers` into the `pwaLikelihood` class, this makes it somewhat simpler to export the function to Python, and also a tiny bit simplifies the code as access to all private members is possible
* add the executable bit, add a warning when overwriting an existing covariance matrix and a check for positive eigenvalues of the Hessian to `calcCovMatrixForFitResult`


legordian:
Merged, thank you!

PR #131, opened 2015-10-19
pyInterface: Improve eigenvectorLikelihoodSlices.

legordian:
Improve the eigenvectorLikelihoodSlices script by adding an option telling the
script not to calculate the Hessian analytically (because it was already
calculated that way when the fit result file was produced). Also fix a problem
with the python version of the '*' operator.

### DIFF
--- a/pyInterface/userInterface/eigenvectorLikelihoodSlices.py
+++ b/pyInterface/userInterface/eigenvectorLikelihoodSlices.py
@@ -19,6 +19,7 @@ if __name__ == "__main__":
 	parser.add_argument("outputFileName", type=str, metavar="outputFile", help="path to output file")
 	parser.add_argument("-c", type=str, metavar="configFileName", dest="configFileName", default="./rootpwa.config", help="path to config file (default: './rootpwa.config')")
 	parser.add_argument("-b", type=int, metavar="#", dest="binID", default=0, help="bin ID of fit (default: 0)")
+	parser.add_argument("-s", action="store_true", dest="ranWithHesse", default=False, help="indicates that the fit result contains an analytically calculated covariance matrix")
 	parser.add_argument("-C", "--cauchyPriors", help="use half-Cauchy priors (default: false)", action="store_true")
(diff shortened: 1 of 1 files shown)


pyInterface/userInterface/eigenvectorLikelihoodSlices.py
```diff
@@ -128,7 +130,9 @@
 		graphLikeli = ROOT.TGraph()
 		for p in xrange(-50, 51):
 			ratio = (p/50.0) * math.sqrt(eigenVectorsMinuit[par][1])
-			pars = minimum + ratio * eigenVectorsMinuit[par][0]
+			step = ROOT.TVectorD(eigenVectorsMinuit[par][0])
+			step *= ratio
+			pars = minimum + step
```
suhlatwork: Could you add a TODO marker here reminding use to properly write this once we all use ROOT versions where this is possible to do?

pyInterface/userInterface/eigenvectorLikelihoodSlices.py
```diff
@@ -78,12 +79,13 @@
...
-	covMatrixAna = likelihood.CovarianceMatrix( [ minimum[i] for i in xrange(minimum.GetNrows()) ] )
-	eigenVectorsAna = likelihood.HessianEigenVectors(covMatrixAna)
+	if not args.ranWithHesse:
+		covMatrixAna = likelihood.CovarianceMatrix( [ minimum[i] for i in xrange(minimum.GetNrows()) ] )
+		eigenVectorsAna = likelihood.HessianEigenVectors(covMatrixAna)
 
-	if len(eigenVectorsMinuit) != len(eigenVectorsMinuit):
-		pyRootPwa.utils.printErr("different number of Eigenvalues for covariance matrix of Minuit vs. analytic covariance matrix.")
-		sys.exit(1)
+		if len(eigenVectorsMinuit) != len(eigenVectorsMinuit):
```
suhlatwork: This comparison should be fixed, we should comare `eigenVectorsMinuit` and `eigenVectorsAna` here. Don't know why I didn't write this correctly in the first place, ...

PR #134, opened 2015-12-21
pwaLikelihood: support boost and ROOT wrappers

fkrinner:
add a second addNormIntegral and addAccIntegral python wrapper
to also support boost wrapped ampIntegralMatrices alongside with
the ROOT wrapped ones


legordian:
Merged, thank you!

PR #136, opened 2016-01-19
Fix and improve resonance fit.

suhlatwork:
* Fix a bug while using branchings, the cache was not invalidated correctly. Fits using only couplings were not affected.
* Fix a bug while writing the result of a fit using branchings to the output file. Fits using only couplings were not affected.

* No longer automatically free the couplings, they have to be explicitly freed using the `freeparameters` option. The default if the `freeparameters` option is not in the option file is to free the couplings.
* Create the graphs of the fit result with an adjustable granularity, no longer use the smooth line (ROOT TGraph draw option `C`) but a simple polyline (`L`). The default is to only have points at the mass bin centers, with the command line option `-F` the number of points can be increased (default value 1).
* In case multiple bins are fitted, also create graphs of the intensities showing the sum of all bins.

* Refactor plotting of final-state mass-dependence.
* Remove draw options that have no effect.
* More cosmetic changes.


legordian:
Merged, thank you!

PR #137, opened 2016-01-19
Try to get travis to work again.

suhlatwork:
`wget` apparently does not understand the new certificate use for the `Cmake` servers, so use `curl` instead.

The error message was
```
ERROR: no certificate subject alternative name matches
	requested host name `cmake.org'.
```


legordian:
Merged, thank you!

PR #139, opened 2016-01-22
importanceSampling added

fkrinner:


suhlatwork:
* does not compile without BAT (as Travis tries to at the moment)
* as you are using the BAT interfaces from version 1.0 you should increase the required version, as in 8c91021bc3abb438661dc97415c2cbbd18506815

suhlatwork:
* does not compile with BAT (as you apparently nowhere define that the BAT include dirs are to be used)
* does not link if BAT is configured with the `--enable-parallel` option
* could you include a python script making use of this new functionality?

suhlatwork:
* the new `genPseudoDataImportanceSampling.py` cannot be used to create events at a fixed mass `-B 0`
* compared to `generatePhaseSpace` some options are missing: `--beamfile <file>`, `--noRandomBeam`, `--randomBlockBeam`(to be fair those are also missing in `genPseudoData.py`)
* running twice with the same seed will give different results
* this looks wrong wrongmass probably the used formula to calculate the break-up momenta is not suited for this purpose
* are the squares (`^2`) missing for the axis labels of some of the BAT plots
* probably all binning variables should be stored in the event meta data (and not just the mass), otherwise the pseudo data will not be compatible with the rest of the data from the file managers point-of-view
* the tPrime distribution is wrong, tPrime calculated from the momentum vectors does not agree with what is stored in the additional variable wrongtprime (should be limited between 0.1 and 1 to state the most obvious problem)
* for the pi-pi0pi0 final state the generated events do not seem to be flat in phase-space
 * the 2D plots of the angles in the helicity frame have stripes wrongangles
 * the energies of the outgoing particles have wiggles wronge

suhlatwork:
Further issues:
* Using `clang` the changes no longer compile due to the unused variable `nChainsDefault`.
* Why do you need the `rpwa::importanceSampler::SetNChains(unsigned)` method? Boost.Python is able to export inherited methods.
* Why do you have your own formula to calculate the break-up momentum and do not use the one from `utilities/physUtils.hpp`?
* Correct the spelling of `phaseSpaceOnly`/`phaseSpeaceOnly` in the Python script.
* `generator.overrideBeamFile` needs to be called before `generator.readReactionFile`.
* The Python script does not work with the default value for `lag`, as it then sets the number of iterations to zero, which is not caught by BAT.
* Stick to @legordian's coding style.

A more general issue:
* Should the generator config file be added to `rootPwaConfig`?

pyInterface/genPseudoDataImportanceSampling.py
```diff
@@ -0,0 +1,184 @@
...
+#	modelSampler.setPhaseSpaceOnly()
+	modelOutput =  pyRootPwa.ROOT.TFile.Open(args.outputFile, "NEW")
+	modelSampler.initializeFileWriter(modelOutput)
+	modelSampler.SetNChains(args.nChains)
+	modelSampler.SetNIterationsRun(args.nEvents/args.nChains*args.lag)
+	modelSampler.SetNLag(args.lag)
+	if args.phaseSpeaceOnly:
+		modelSampler.setPhaseSpaceOnly()
+
+#	pars = [.7,.7,.7,.7,.7, 1.8]
```
fkrinner: remove this

suhlatwork:
I started working on this again, so far the status from commit 0f100e2fe85709dd07f2e4429b73503d587224ad has been rebased onto the current master and coding-style issues have been fixed. The current progress can be tracked in my `importanceSampling` branch.

The issues encountered so far are:
* it does not work properly if BAT has been compiled with support for OpenMP unless `OMP_NUM_THREADS=1` is set as an environment variable (note that this is probably not a problem of the changes themselves, but improper usage of the amplitude calculation)
* the python script finishes with a segmentation violation
* the `modelIntensity` class contains the fourth (or so) attempt to sum up amplitudes correctly (`pwaLikelihood`, `fitResult`, `weightEvents` are those that come to my mind immediately). While there is a clear distinction to what is done in `pwaLikelihood`, one should be able to implement a (rather slim) method covering all the other use cases (probably in `fitResult`). (also at the moment it is only working for rank-1.)
* there is quite some code duplication for the calculation of the four-vectors

I will try to work on those issues in the next days.

suhlatwork:
* in contrast to `genPseudoData.py` the mass distribution within the given mass range is not flat, but follows phase space.

PR #141, opened 2016-03-03
Update README.

bgrube:
Update and correct installation instructions.


README.md
```diff
@@ -181,7 +198,7 @@ The command will download the code into the `ROOTPWA` directory. This working co
 
 ROOTPWA is developed in multiple code branches, most of which contain work in progress that might not work as expected. However, there are dedicated stable branches that are intended to be used for real analyses. Tested versions of these branches are identified by tags.
 
-By default you will be in the `master` branch after you cloned the repository. This branch is used for development and will contain the latest features. More conservative users are recommended to use a more stable branch. At the moment the latest stable branch is called `_v2`. You may also checkout tagged versions. The list of tags is given by
+By default you will be in the `master` branch after you cloned the repository. This branch is used for development and will contain the latest features. More conservative users are recommended to use stable branches. Note, however, that new features will not be backported to stable branches. At the moment the latest stable branch is called `_v2`. You may also checkout tagged versions. The list of tags is given by
```
suhlatwork: Should we update the latest stable branch to at least `_v3`, or even to some point after the inclusion of the new Python user interface? With `_v3` we at least would have the same directory structure and library names as we have now in `master`. But even the amplitude files created with `_v3` cannot be used with the Python interface, and old integral files would at least have to be renamed according to the naming conventions of the file manager. The difference to branches we previously called stable branches is that we already know `_v3` is a dead end.

Issue #140, opened 2016-03-01
fitResult::intensityErr(const std::string&) broken for non-rank-1 fits


suhlatwork:
In fitResult.cc starting from line 571

```
	// get amplitudes that correspond to wave name pattern
	const vector<unsigned int> prodAmpIndices = prodAmpIndicesMatchingPattern(waveNamePattern);
	const unsigned int         nmbAmps        = prodAmpIndices.size();
	if (not _covMatrixValid or (nmbAmps == 0)) {
		return 0.;
	}
	if(nmbAmps == 1) {
		return intensityErr(prodAmpIndices[0]);
	}
```

This will get the indices to the production amplitudes based on the pattern specified, and if only one production amplitude is found, call `intensityErr` with the corresponding index. However, `intensityErr` takes wave indices. The indices for production amplitudes and waves are different for anything but rank-1 fits. Symptoms might vary from a wrong value to crashes/exceptions.

This was caused by a change made in a4e9e17.

Issue #60, opened 2014-12-02
Arguments to intensity/intensityErr


suhlatwork:
The arguments of the `fitResult::intensity` method are regular expressions for wave names, the arguments of the `fitResult::intensityErr` method are regular expressions for production amplitudes. Usually this is not a big problem, but if one requires that the pattern has to match from the beginning of the wave name (i.e. it starts with the ^ character), then intensityErr will not return a valid result. This should somehow be improved.

It seems that at the moment it is not possible to obtain the intensity in one particular rank (but only the sum over all ranks), so probably the way out is to use a pattern for the production amplitudes also in `fitResult::intensity`.

PR #149, opened 2016-04-28
fitResult: Fix fitResult::intensityErr(const std::string&) broken for non-rank-1 fits

stefan-wallner:
Pull request which fixes issue #140 according to the instructions of @suhlatwork .

### DIFF
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -569,14 +569,21 @@ double
 fitResult::intensityErr(const string& waveNamePattern) const
 {
 	// get amplitudes that correspond to wave name pattern
-	const vector<unsigned int> prodAmpIndices = prodAmpIndicesMatchingPattern(waveNamePattern);
-	const unsigned int         nmbAmps        = prodAmpIndices.size();
-	if (not _covMatrixValid or (nmbAmps == 0)) {
+	const vector<unsigned int> waveIndices    = waveIndicesMatchingPattern(waveNamePattern);
+	const unsigned int         nmbWaves       = waveIndices.size();
+	if (not _covMatrixValid or (nmbWaves == 0)) {
 		return 0.;
 	}
-	if(nmbAmps == 1) {
-		return intensityErr(prodAmpIndices[0]);
+	if(nmbWaves == 1) {
+		return intensityErr(waveIndices[0]);
 	}
+
+	vector<unsigned int> prodAmpIndices;
+	for (unsigned int i = 0; i < nmbWaves; ++i) {
+		const vector<unsigned int> prodAmpIndicesWave = prodAmpIndicesForWave(waveIndices[i]);
+		prodAmpIndices.insert(prodAmpIndices.end(), prodAmpIndicesWave.begin(), prodAmpIndicesWave.end());
+	}
+	const unsigned int nmbAmps = prodAmpIndices.size();
 	// build Jacobian for intensity, which is a 1 x 2n matrix composed of n sub-Jacobians:
 	// J = (JA_0, ..., JA_{n - 1}), where n is the number of production amplitudes
 	TMatrixT<double> jacobian(1, 2 * nmbAmps);


partialWaveFit/fitResult.cc
```diff
@@ -519,9 +519,7 @@ double
 fitResult::intensity(const string& waveNamePattern) const
 {
 	vector<unsigned int> waveIndices = waveIndicesMatchingPattern(waveNamePattern);
-	if(waveIndices.size() == 1) {
-		return intensity(waveIndices[0]);
-	}
+
```
suhlatwork: Contrary to what I previously said, this appears correct to me now and should not be changed.

stefan-wallner: Should both functions not be changed?

partialWaveFit/fitResult.cc
```diff
@@ -574,9 +572,6 @@ fitResult::intensityErr(const string& waveNamePattern) const
 	if (not _covMatrixValid or (nmbAmps == 0)) {
 		return 0.;
 	}
-	if(nmbAmps == 1) {
-		return intensityErr(prodAmpIndices[0]);
-	}
```
suhlatwork: Here some fixing is indeed required, but I am not sure that removing those lines (and thus reverting a4e9e17cbd055ec9a0f2793bdac5dcf083626da4) is a good choice. Initially commit a4e9e17cbd055ec9a0f2793bdac5dcf083626da4 was made to allow plotting intensities (and all other information of a single spin-density matrix element) with errors also for fit results without integral matrices. This functionality needs to be kept.

stefan-wallner:
Readded statement to the intensity function.
Implemented intensityErr function in a way that it SHOULD work with rank-2 and still has the functionality that intensity errors of waves can be calculated even without integral matrix.

partialWaveFit/fitResult.cc
```diff
@@ -569,14 +569,17 @@ double
...
+	if (not _covMatrixValid or (nmbWaves == 0)) {
 		return 0.;
 	}
-	if(nmbAmps == 1) {
-		return intensityErr(prodAmpIndices[0]);
+	if(nmbWaves == 1) {
+		return intensityErr(waveIndices[0]);
 	}
+
+	const vector<unsigned int> prodAmpIndices = prodAmpIndicesMatchingPattern(waveNamePattern);
```
suhlatwork: `waveNamePattern` is now used once as a pattern for a wave name and once for a pattern for production amplitudes (the name `waveNamePattern` obviously was misleadingly used to begin with). When creating the standard spin totals for three pions I observed the necessity to prefix the pattern for the production amplitudes with `V\\d*_` as otherwise wrong patterns would be matched and the intensity error would be calculated wrong. From a performance point of view it would be a nightmare to write something like:
```
vector<unsigned int> prodAmpIndices;
for (unsigned int i = 0; i < nmbWaves; ++i) {
	const vector<unsigned int> prodAmpIndicesWave = prodAmpIndicesForWave(waveIndices[i]);
	prodAmpIndices.insert(prodAmpIndices.end(), prodAmpIndicesWave.begin(), prodAmpIndicesWave.end());
}
```

suhlatwork:
Relevant for #60 and #140.

partialWaveFit/fitResult.cc
legordian: Without changing the structure of the fitResult class, I think @suhlatwork's solution is the only viable one which fixes both #60 and #140. I have no feeling for how big the performance hit will be, but to extract the correct uncertainty on the intensity, one has to do a loop like this anyway somewhere.

stefan-wallner: As far as I understand the code, 'prodAmpIndicesForWave()' resolves the amplitude ids by the corresponding wave name, which is of course computationally quite expensive. 
However, when calling an element of the spin-density matrix, the 'prodAmpIndicesForWave()' is called anyway to resolve the mapping between amplitude ids and wave ids. 
Is there no direct mapping between the amplitude ids and the wave id, because I thought for rank=1 fits they are the same. 
Or would it be possible / sensible to construct such a mapping once and store it with the fit result?

stefan-wallner: After a discussion with @legordian and @bgrube, I implemented the solution, suggested by @suhlatwork. 
The performance impact should hopefully not be to large, as the amplitude-indices-wave-index mapping is done in the same way in the spinDensityMatrixElem member functions. 
However, this **changes the interface of the intensityErr member function**, as the argument is no longer an amplitude-name regex but a wave-name regex, which makes it consistent with the intensity member function.

legordian:
Replaced by #151 and therefore closed.

PR #150, opened 2016-05-02
Fix disabling of symmetrizations.

legordian:
Fix the disabling of the Bose- and Isospin symmetrization mechanisms, which
were just left with no amplitude terms at all until now.


suhlatwork:
Merged. Thanks to @fkrinner and @legordian!

PR #151, opened 2016-05-06
Fix calculation of intensity errors

### LINKED ISSUE #60: Arguments to intensity/intensityErr

### LINKED ISSUE #140: fitResult::intensityErr(const std::string&) broken for non-rank-1 fits

suhlatwork:
In addition to a crash while calculating the errors for a rank-2 fit (#149), there is a much more serious bug in `fitResult::intensityErr`. The function call to `normIntegralForProdAmp` should not use the indices of the `prodAmpIndices` array `i` and `j`, but the content of that array at the respective positions. Patterns that match a single wave should be fine. Patterns that match multiple waves probably return a wrong error for all patterns that did not result in a continuous sequences of production amplitude indices and were not starting from production amplitude index zero (realistically that is all except the '.*' pattern).

The actual impact of this bug is fortunately hardly visible in the plots I compared so far.

Replaces #149.
Fixes #140.
Fixes #60.

### DIFF
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -595,7 +595,7 @@ fitResult::intensityErr(const string& waveNamePattern) const
 		for (unsigned int j = 0; j < nmbAmps; ++j) {
 			if (rankOfProdAmp(prodAmpIndices[j]) != currentRank)
 				continue;
-			ampNorm += prodAmp(prodAmpIndices[j]) * normIntegralForProdAmp(j, i);  // order of indices is essential
+			ampNorm += prodAmp(prodAmpIndices[j]) * normIntegralForProdAmp(prodAmpIndices[j], prodAmpIndices[i]);  // order of indices is essential
 		}
 		jacobian[0][2 * i    ] = ampNorm.real();
 		jacobian[0][2 * i + 1] = ampNorm.imag();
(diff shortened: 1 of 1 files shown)


legordian:
Merged, thank you!

PR #152, opened 2016-05-06
Fix compilation with latest Xcode

suhlatwork:
Fix compilation of `complexMatrix` with the latest version of Apple's Xcode.


legordian:
So there is no way to multiply two `boost::numeric::ublas::matrix` objects with each other?

legordian:
Merged, thank you!

PR #154, opened 2016-05-12
Fix angle plotting script.

legordian:
Fix the angle plotting script by making use of the metadata and removing the
old logic of finding input files. Instead, the input files have to be given
explicitly. This is an intermediate step, as this script should be transformed
into a function in the pyRootPwa package and then called with a new script
using the file manager to figure out where the input files are.


suhlatwork:
Merged. Thank you!

PR #155, opened 2016-05-13
Fix compilation issues with '-Wextra'.

suhlatwork:
Fix two issues when compiling with '-Wextra'.


legordian:
Merged, thank you!

PR #156, opened 2016-05-17
pyInterface: fix 'getWaveDescThresFromWaveList'

suhlatwork:
The python function 'getWaveDescThresFromWaveList' should check whether
a line read from the wavelist can be split into either one or two parts,
not whether the split array is either '1' or '2'. This was broken by commit
71a39d39057ddc47868cbba46cb12868c9806cb5,


legordian:
Merged, thank you!

PR #157, opened 2016-05-17
ampIntegralMatrixMetadata: small patch

fkrinner:
add dictionaries for all subclasses.
existing amplitudeHash not fatal, since in special cases
the same hases can appear several times.


legordian:
Fine by me.

suhlatwork:
Merged. Thank you!

Note that the changes to the LinkDef would not have been required for ROOT 6, and that for ROOT 5 apparently a couple of rules need to be followed:
* `#pragma` statements used for I/O should not have the `std::` namespace
* for `#pragma` statements that are only used to make a class known to ROOT, the `std::` might be present
* ROOT 5 cannot resolve that `size_t` and `unsigned long` are the same type

PR #159, opened 2016-05-19
Fix 'GenerateKeyfiles.sh' scripts

suhlatwork:
* fix behaviour if `WAVESET_FILES` contains the full paths
* allow to specify template keyfile with `TEMPLATE_KEY_FILES` environment variable, which no longer requires the template keyfiles to be present in the current directory


legordian:
Merged, thank you!

PR #160, opened 2016-05-19
pyInterface: add option to fit without acceptance correction

suhlatwork:
Add a '--noAcceptance' option to 'pwaFit.py' and 'pwaNloptFit.py' which
allows to perform fits without taking the acceptance into account.


legordian:
Merged, thank you!

PR #161, opened 2016-05-19
add createIsobarDecayTopology to py

fkrinner:


pyInterface/decayAmplitude/isobarDecayTopology_py.cc
```diff
@@ -159,6 +159,16 @@ namespace {
 		return bp::list(self.findIsobarBoseSymVertices());
 	}
 
+	rpwa::isobarDecayTopologyPtr createIsobarDecayTopology_1(const rpwa::isobarDecayTopology& topo)
+	{
+		return rpwa::createIsobarDecayTopology(topo);
+	}
+
+	rpwa::isobarDecayTopologyPtr createIsobarDecayTopology_2(const rpwa::decayTopology& topo)
```
suhlatwork: shouldn't this return `rpwa::decayTopologyPtr`?

pyInterface/decayAmplitude/isobarDecayTopology_py.cc
```diff
@@ -238,4 +248,13 @@ void rpwa::py::exportIsobarDecayTopology() {
...
 	bp::register_ptr_to_python<rpwa::isobarDecayTopologyPtr>();
 
+	bp::def(
+		"createIsobarDecayTopology"
+		, &createIsobarDecayTopology_1
+	);
+	bp::def(
+		"createIsobarDecayTopology"
+		, &createIsobarDecayTopology_2
+	);
```
suhlatwork: Is this really the correct order to specify the functions, I seem to remember that the first function defined is the last function that is tried, so the first defined function should be the more general. Did you make sure that the correct functions are called for both arguments?

suhlatwork:
Merged by hand (58af4231eb7ee5c73197c0164abc71780fd4de91) after solving the remaining issue.

PR #162, opened 2016-05-20
buildSystem: linker flags and new version for BAT

suhlatwork:
Require at least BAT version 1.0. There was a massive interface change
with this version so we cannot support earlier and later versions at the
same time, so we will only support the new versions.

Remove leading and trailing whitespaces from BAT's special linker flags.


legordian:
Merged, thank you!

PR #163, opened 2016-05-20
Split 'nBodyPhaseSpaceGen' class

suhlatwork:
Move those parts of `nBodyPhaseSpaceGen` that calculate weights and kinematic variables into a new class `nBodyPhaseSpaceKinematics`. Rename `nBodyPhaseSpaceGen` to `nBodyPhaseSpaceGenerator`, which now inherits from `nBodyPhaseSpaceKinematics` and still generates the intermediate two-body masses and decay angles. Usage of `nBodyPhaseSpaceGen` can simply be replaced by using `nBodyPhaseSpaceGenerator`.

This should allow to use use the calculation of weight and momenta more simply with other sources for intermediate masses and angles. This commit is the first part of an attempt to reduce the code duplication of #139.


legordian:
Merged, thank you!

PR #164, opened 2016-05-23
Revise the file manager for on-the-fly binning.

legordian:
Revise the file manager and all the attached scripts to implement a binning logic including on-the-fly binning. At the present stage, this is in a rather rough state: all the scripts should work, but the functionality has not been fully exploited yet. In particular, the fitting scripts in this version are supplied with a integer bin id, which corresponds to an integral bin. This should be replaced by an interface actually passing in the bin and the script system using the integral bins to create an new integral for that bin, if possible. Also, a number of checks are not implemented yet, e.g. it is not checked if every event is actually in a bin and the metadata compatibility is not verified. However, this changeset should provide a good basis to implement these things.


pyInterface/package/utils/_binning.py
```diff
@@ -0,0 +1,106 @@
...
+			if binRange[0] >= binRange[1]:
+				raise ValueError("Lower bound of bin range (" + str(binRange[0]) + ") is larger or equal to upper bound of bin range (" + str(binRange[1]) + ").")
+			if isinstance(binRange[0], int) or isinstance(binRange[1], int):
+				boundaries[key] = (float(binRange[0]), float(binRange[1]))
+		self.boundaries = boundaries
+
+
+	def __lt__(self, other):
+		keys = sorted(self.boundaries.keys())
+		if keys != sorted(self.boundaries.keys()):
```
fkrinner: This comparison seems to be true always. Shouldn't you compare with other.boundaries.keys()?

pyInterface/package/utils/_waveDescThresUtils.py
```diff
@@ -49,3 +49,30 @@ def getWaveDescThresFromWaveList(waveListFileName, keyFiles):
...
+def getWaveThresFromWaveList(waveListFileName, waveDescriptions):
+	printInfo("reading amplitude names and thresholds from wave list file "
+	          + "'" + waveListFileName + "'.")
+	waveDescThres = []
+	with open(waveListFileName, 'r') as waveListFile:
+		lineNmb = 0
+		for line in waveListFile:
+			if line[0] == '#':  # comments start with #
+				continue
+			line = line.replace('\n', '')
```
fkrinner: Couldn't you use line.strip()? If you do this before the check for comments, lines starting with " #" would also be recognized as comment

pyInterface/package/_fileManager.py
```diff
@@ -108,187 +136,144 @@ def initialize(self, configObject):
...
-				if not self.binList[binID][binningVariable] == dataFile.binningMap[binningVariable]:
-					found = False
-					break
-			if found:
-				return dataFile
-		pyRootPwa.utils.printWarn("no data dataFile found for binID = " + str(binID) + "and eventsType = '" + str(eventsType) + "'.")
-		return None
-
+	def checkBinning(self):
+		raise Exception("not implemented. " + repr(self))
```
fkrinner: Maybe `raise NotImplementedError("...")`?

decayAmplitude/ampIntegralMatrix.cc
```diff
@@ -296,9 +297,11 @@ ampIntegralMatrix::addEvent(map<string, complex<double> > &amplitudes)
 }
 
 bool
-ampIntegralMatrix::integrate(const vector<const amplitudeMetadata*>& ampMetadata,
-                             const unsigned long                     maxNmbEvents,
-                             const string&                           weightFileName)
+ampIntegralMatrix::integrate(const vector<const amplitudeMetadata*>&   ampMetadata,
+                             const unsigned long                       maxNmbEvents,
+                             const string&                             weightFileName,
+                             const rpwa::eventMetadata*                eventMeta,
```
suhlatwork: you removed the `rpwa::` above, so it should probably be removed here as well

decayAmplitude/ampIntegralMatrix.cc
```diff
@@ -378,9 +416,25 @@ ampIntegralMatrix::integrate(const vector<const amplitudeMetadata*>& ampMetadata
 	vector<vector<complex<double> > > amps(_nmbWaves);
 	progress_display progressIndicator(_nmbEvents, cout, "");
 	bool             success = true;
+	unsigned long eventCounter = 0;
```
suhlatwork: `eventCounter` and `success` should be aligned

decayAmplitude/ampIntegralMatrix.cc
```diff
@@ -370,6 +373,41 @@ ampIntegralMatrix::integrate(const vector<const amplitudeMetadata*>& ampMetadata
 		useWeight = true;
 	}
 
+	TTree* eventTree = 0;
+	vector<double> binningVariables(otfBin.size());
+	vector<pair<double, double> > bounds(otfBin.size());
+	if(eventMeta) {
+		if(otfBin.empty()) {
+			printErr << "got event metadata but the binning map is emtpy." << endl;
```
suhlatwork: why is this a problem?

decayAmplitude/ampIntegralMatrix.cc
```diff
@@ -370,6 +373,41 @@ ampIntegralMatrix::integrate(const vector<const amplitudeMetadata*>& ampMetadata
...
+			          << elem->second.second << "]'." << endl;
+			int err = eventTree->SetBranchAddress(elem->first.c_str(), &binningVariables[otfBinIndex]);
+			bounds[otfBinIndex] = elem->second;
+			++otfBinIndex;
+			if(err < 0) {
+				printErr << "could not set branch address for branch '" << elem->first << "' (error code " << err << ")." << endl;
+				return false;
+			}
+		}
+		if(binningVariables.size() != bounds.size()) {
```
suhlatwork: this comparison is always false

decayAmplitude/ampIntegralMatrixMetadata.cc
```diff
@@ -228,48 +222,12 @@ bool rpwa::ampIntegralMatrixMetadata::setAllZeroHash() {
 }
 
 
-bool rpwa::ampIntegralMatrixMetadata::addEventMetadata(const rpwa::eventMetadata& evtMeta, size_t eventMin, size_t eventMax) {
-	if (eventMin >= eventMax) {
-		printErr << "given event number bounds are not ordered correctly." << endl;
+bool rpwa::ampIntegralMatrixMetadata::addEventMetadata(const rpwa::eventMetadata& evtMeta) {
+	if (std::find(_evtMetas.begin(), _evtMetas.end(), evtMeta) != _evtMetas.end()) {
+		printWarn << "event metadata object has already been added." << endl;
```
suhlatwork: this is too prohibitive, the same event metadata might appear several times if the binning of integrals and events does not match

decayAmplitude/ampIntegralMatrix.h
```diff
@@ -122,11 +123,11 @@ namespace rpwa {
...
-
-
-		bool integrate(const std::vector<const rpwa::amplitudeMetadata*>& ampMetadata,
-		               const unsigned long                                maxNmbEvents   = 0,
-		               const std::string&                                 weightFileName = "");
+		bool integrate(const std::vector<const rpwa::amplitudeMetadata*>&       ampMetadata,
+		               const unsigned long                                      maxNmbEvents   = 0,
+		               const std::string&                                       weightFileName = "",
+		               const rpwa::eventMetadata*                               eventMeta = 0,
+		               const std::map<std::string, std::pair<double, double> >& otfBin = std::map<std::string, std::pair<double, double> >());
```
suhlatwork: align the `=` signs?

pyInterface/deWeight.py
```diff
@@ -53,19 +64,25 @@
 	                             metaData.productionKinematicsParticleNames(),
 	                             metaData.decayKinematicsParticleNames(),
 	                             metaData.binningMap(),
-	                             ["weight"]):
+	                             additionalVariableLabels):
```
suhlatwork: the weight should not be passed on, after the deweighting it is 1

decayAmplitude/ampIntegralMatrix.cc
suhlatwork: the other way around would be a problem

pyInterface/package/_amplitude.py
```diff
@@ -7,7 +7,7 @@
 
 def calcAmplitude(inputFileName,
                   keyFileName,
```
suhlatwork: `keyFileName` could be replaced by `waveName` as the key file is no longer used in this function

pyInterface/package/_amplitude.py
```diff
@@ -43,10 +43,9 @@ def calcAmplitude(inputFileName,
 		printWarn("key file '" + keyFileName + "' not found.")
 		outputFile.Close()
 		return False
```
suhlatwork: this check can be removed

pyInterface/package/_config.py
```diff
@@ -94,3 +101,113 @@ def initialize(self, configFileName):
...
+				pyRootPwa.utils.printWarn(errMsg)
+				raise ValueError(errMsg)
+			expandedBoundaryList = []
+			for i in xrange(axesPartition[2]+1):
+				expandedBoundaryList.append(float(axesPartition[0]) + (float(axesPartition[1] - axesPartition[0]) / float(axesPartition[2])) * i)
+			inputVal[key] = expandedBoundaryList
+	return inputVal
+
+
+def _testReadBinning():
```
suhlatwork: move to `testPyRootPwa.py`?

pyInterface/package/_fileManager.py
```diff
@@ -81,21 +92,38 @@ def initialize(self, configObject):
...
+		if not self.binList:
+			for _, inputFiles in self.dataFiles.iteritems():
+				for inputFile in inputFiles:
+					pyRootPwa.utils.printInfo("adding bin '" + str(inputFile.binningMap) + "'.")
+					try:
+						self.binList.append(pyRootPwa.utils.multiBin(inputFile.binningMap))
+					except (TypeError, ValueError):
+						pyRootPwa.utils.printErr("no binning given in config file and no binning map" +
+						                         " found in data file '" + str(inputFile.dataFileName) + "'.")
+						return False
```
suhlatwork: check non-overlapping bins?

pyInterface/package/_fileManager.py
suhlatwork: or simply return `True` for the moment, but call the function at the appropriate places

pyInterface/package/_fileManager.py
```diff
@@ -33,42 +32,54 @@ def loadFileManager(path):
...
+		retval = ""
+		retval += "'" + self.dataFileName + "': ("
+		retval += str(self.binningMap) + ", "
+		retval += str(self.eventsType) + ", "
+		retval += str(self.additionalVariables) + ")"
+		return retval
 
 
-class EventsType:
+class EventsType(object):
```
suhlatwork: can't we use the exported enum from `rpwa::eventMetadata`? that way also the conversion function could be removed.

pyInterface/package/_fileManager.py
```diff
@@ -108,187 +136,144 @@ def initialize(self, configObject):
...
+		if waveName not in self.keyFiles:
+			pyRootPwa.utils.printErr("wave name '" + str(waveName) + "' cannot be produced by any of the keyfiles.")
+			return None
+		keyFileName = self.keyFiles[waveName]
+		waveDescriptions = pyRootPwa.core.waveDescription.parseKeyFile(keyFileName)
+		for waveDescriptionID, waveDescription in enumerate(waveDescriptions):
+			(success, amplitude) = waveDescription.constructAmplitude()
+			if not success:
+				pyRootPwa.utils.printErr("could not construct decay topology for wave descrption at index " + str(waveDescriptionID) + " of key file '" + keyFileName + "'.")
+				return collections.OrderedDict()
```
suhlatwork: return `None`?

pyInterface/package/_fileManager.py
```diff
@@ -108,187 +136,144 @@ def initialize(self, configObject):
...
-		return self.keyFiles
+	def getEventAndAmplitudePairPathsForWave(self, eventsType, waveName):
+		eventsType = fileManager.pyEventsType(eventsType)
+		retval = []
+		if eventsType not in self.dataFiles:
+			pyRootPwa.utils.printWarn("events type '" + str(eventsType) + "' not found.")
+			return []
+		if waveName not in self.keyFiles:
+			pyRootPwa.utils.printWarn("key file for wave name '" + waveName + "' not found.")
+			return []
```
suhlatwork: here and above return `None`?

pyInterface/package/_fileManager.py
```diff
@@ -108,187 +136,144 @@ def initialize(self, configObject):
...
+		if eventsType not in self.dataFiles:
+			pyRootPwa.utils.printWarn("events type '" + str(eventsType) + "' not in data files.")
+			return []
+		eventFileIds = []
+		for eventFileId, inputFile in enumerate(self.dataFiles[eventsType]):
+			found = True
+			for variableName in inputFile.binningMap:
+				if variableName in multiBin.boundaries:
+					if (inputFile.binningMap[variableName][1] < multiBin.boundaries[variableName][0]) or \
+					   (inputFile.binningMap[variableName][0] > multiBin.boundaries[variableName][1]):
```
suhlatwork: should this be `>=`?

pyInterface/package/_fileManager.py
```diff
@@ -108,187 +136,144 @@ def initialize(self, configObject):
...
+			for inputFile_i, _ in enumerate(self.dataFiles[eventsType]):
+				for waveName_i, waveName in enumerate(self.keyFiles.keys()):
+					amplitudeFileName = waveName + "_eventFileId-" + str(inputFile_i) + "_" + str(eventsType) + ".root"
+					if needSubdirs:
 						subDir = fileCount/self.limitFilesInDir
 						fullDir = self.amplitudeDirectory + "/" + str(subDir)
 						if not fullDir in dirSet:
 							dirSet.append(fullDir)
-						amplitudeFiles[(binID, waveName, eventsType)] = str(subDir) + "/" + waveName + "_binID-" + str(binID) + "_" + str(eventsType) + ".root"
+						amplitudeFileName = str(subDir) + "/" + amplitudeFileName
```
suhlatwork: replace `str(subDir)` by `fullDir`?

pyInterface/package/_fileManager.py
```diff
@@ -304,102 +289,100 @@ def _getAmplitudeFilePaths(self):
...
-				return []
-			for waveDescriptionID in xrange(len(waveDescriptions)):
-				waveDescription = waveDescriptions[waveDescriptionID]
+				return collections.OrderedDict()
+			for waveDescriptionID, waveDescription in enumerate(waveDescriptions):
 				(success, amplitude) = waveDescription.constructAmplitude()
 				if not success:
 					pyRootPwa.utils.printErr("could not construct decay topology for wave descrption at index " + str(waveDescriptionID) + " of key file '" + keyFileName + "'.")
-					return []
+					return collections.OrderedDict()
```
suhlatwork: the usual return value seems to be a standard dictionary, vie return an `OrderedDict` in case of error?

suhlatwork: nevermind, ...

pyInterface/package/_likelihood.py
```diff
@@ -40,18 +41,34 @@ def initLikelihood(waveDescThres,
...
+		eventFile = ROOT.TFile.Open(eventFileName, "READ")
+		if not eventFile:
+			pyRootPwa.utils.printErr("could not open event file '" + eventFileName + "'. Aborting...")
+			return False
+		eventMeta = pyRootPwa.core.eventMetadata.readEventFile(eventFile)
+		if not eventMeta:
+			pyRootPwa.utils.printErr("could not read metadata from event file '" + eventFileName + "'. Aborting...")
+			return False
+		eventMetas.append(eventMeta)
+	print(multiBin.boundaries)
```
suhlatwork: debug leftover?

pyInterface/package/utils/_binning.py
```diff
@@ -0,0 +1,106 @@
...
+		# binningInfo = { "variableName": value }
+		if sorted(binningInfo.keys()) != sorted(self.boundaries.keys()):
+			return False
+		for variableName, value in binningInfo.iteritems():
+			if value < self.boundaries[variableName][0] or value >= self.boundaries[variableName][1]:
+				return False
+		return True
+
+
+def _testMultiBin():
```
suhlatwork: move to `testPyRootPwa.py`?

pyInterface/package/utils/_waveDescThresUtils.py
```diff
@@ -49,3 +49,30 @@ def getWaveDescThresFromWaveList(waveListFileName, keyFiles):
 			lineNmb += 1
 		printInfo("read " + str(lineNmb) + " lines from wave list file " + "'" + waveListFileName + "'")
 	return waveDescThres
+
+
+def getWaveThresFromWaveList(waveListFileName, waveDescriptions):
```
suhlatwork: why not rename to `getWaveDescThresFromWaveList`? the current `getWaveDescThresFromWaveList` is in any case not usable anymore as the `keyFiles` structure has changed.

pyInterface/userInterface/pwaFit.py
```diff
@@ -47,17 +47,23 @@
...
-	ampFileList = fileManager.getAmplitudeFilePaths(args.binID, pyRootPwa.core.eventMetadata.REAL)
-	if not ampFileList:
+	if args.integralBin < 0:
+		pyRootPwa.utils.printErr("bin < 0 (" + str(args.integralBin) + "). Aborting...")
+		sys.exit(1)
+	elif args.integralBin >= len(fileManager.binList):
+		pyRootPwa.utils.printErr("bin out of range (" + str(args.integralBin) + ">=" + str(len(fileManager.binList)) + "). Aborting...")
+		sys.exit(1)
+	multiBin = fileManager.binList[args.integralBin]
+	massBinCenter = (multiBin.boundaries['mass'][1] + multiBin.boundaries['mass'][0]) / 2.
```
suhlatwork: is `massBinCenter` used anywhere in this script?

pyInterface/userInterface/pwaNloptFit.py
```diff
@@ -47,17 +47,23 @@
...
-	ampFileList = fileManager.getAmplitudeFilePaths(args.binID, pyRootPwa.core.eventMetadata.REAL)
-	if not ampFileList:
+	if args.integralBin < 0:
+		pyRootPwa.utils.printErr("bin < 0 (" + str(args.integralBin) + "). Aborting...")
+		sys.exit(1)
+	elif args.integralBin >= len(fileManager.binList):
+		pyRootPwa.utils.printErr("bin out of range (" + str(args.integralBin) + ">=" + str(len(fileManager.binList)) + "). Aborting...")
+		sys.exit(1)
+	multiBin = fileManager.binList[args.integralBin]
+	massBinCenter = (multiBin.boundaries['mass'][1] + multiBin.boundaries['mass'][0]) / 2.
```
suhlatwork: is `massBinCenter` used anywhere in this script?

suhlatwork:
* the conversion of MeV to GeV also needs to be done for:
  * userAnalysisWorkspace/3pi.--+/keyfiles/wavelist.compass.2004
  * userAnalysisWorkspace/3pi.--+/keyfiles/wavelist.compass.2008.42waves
  * userAnalysisWorkspace/3pi.--+/keyfiles/wavelist.compass.2008.53waves
  * userAnalysisWorkspace/3pi.-00/keyfiles/wavelist.compass.2004
  * userAnalysisWorkspace/3pi.-00/keyfiles/wavelist.compass.2008.42waves
  * userAnalysisWorkspace/3pi.-00/keyfiles/wavelist.compass.2008.53waves
  * userAnalysisWorkspace/3pi.-00/keyfiles/wavelist.compass.2008.88waves
  * storageFormats/convertEventFile.py (argument help text, example for `--binning`)
  * genPseudoData.py (should not multiply by 1000 to get the mass to search the fit result)
  * partialWaveFit/pwaLikelihood.cc (adjust messages showing thresholds)
  * highLevelInterface/pwaNloptFit.cc (adjust message showing mass range of fit)
  * highLevelInterface/pwaFit.cc (adjust message showing mass range of fit)

* my comments are based on commit 06d6d7a7d86a5fe039866fe5c1f8fef47bdb662b
* no actual testing done so far

decayAmplitude/ampIntegralMatrix.cc
legordian: Because both `eventMeta` and `otfBin` have default argument, so if one only supplies the first argument of the two and leaves the second on default, this is an error.

A better way to do this might be passing in a `pair<eventMeta*, otfBin>`, would you agree?

decayAmplitude/ampIntegralMatrixMetadata.cc
legordian: I agree, but it does not make much sense to add the same metadata twice (IMHO), it is already there, is it not? To really keep track of what is in the integral file, the metadata have to be save together with the otfBin, but this is beyond the scope of this work. This issue (and several others in the integral metadata) should be revisited when the on-the-fly integrator is decorated with a script.

pyInterface/package/_config.py
legordian: No. Could be deleted if you do not like it.

pyInterface/package/_fileManager.py
legordian: No, they cannot be pickled.

pyInterface/package/_fileManager.py
legordian: Not so easy, the file binning could be multi-dimensional as well.

pyInterface/package/_fileManager.py
legordian: I removed the function.

pyInterface/package/_fileManager.py
legordian: I like it better if the return type of a function is consistent if possible.

resonanceFit/massDepFit.cc
```diff
@@ -1413,7 +1413,7 @@ rpwa::massDepFit::massDepFit::checkFitResultMassBins(TTree* tree,
 		bool found = false;
 		size_t idxMass=0;
 		while(idxMass<_nrMassBins) {
-			if(abs(_massBinCenters[idxMass]-mass) < 1000.*std::numeric_limits<double>::epsilon()) {
+			if(abs(_massBinCenters[idxMass]-mass) < std::numeric_limits<double>::epsilon()) {
```
suhlatwork: this factor 1000 was no GeV to MeV conversion

resonanceFit/massDepFit.cc
```diff
@@ -1479,15 +1479,15 @@ rpwa::massDepFit::massDepFit::readFitResultMassBins(TTree* tree,
...
+		const double newMass = fit->massBinCenter();
 
 		if(_debug) {
 			printDebug << "entry " << idx << ": center of mass bin at " << newMass << " GeV/c^2" << std::endl;
 		}
 
 		bool found = false;
 		for(size_t idxMass=0; idxMass<_massBinCenters.size(); ++idxMass) {
-			if(abs(_massBinCenters[idxMass]-newMass) < 1000.*std::numeric_limits<double>::epsilon()) {
+			if(abs(_massBinCenters[idxMass]-newMass) < std::numeric_limits<double>::epsilon()) {
```
suhlatwork: this factor 1000 was no GeV to MeV conversion

resonanceFit/massDepFit.cc
```diff
@@ -1511,7 +1511,7 @@ rpwa::massDepFit::massDepFit::readFitResultMassBins(TTree* tree,
 
 	_massStep = (_massBinCenters[_nrMassBins - 1] - _massBinCenters[0]) / (_nrMassBins - 1);
 	for(size_t idxMass=1; idxMass<_nrMassBins; ++idxMass) {
-		if(abs(_massBinCenters[idxMass]-_massBinCenters[idxMass-1] - _massStep) > 1000.*std::numeric_limits<double>::epsilon()) {
+		if(abs(_massBinCenters[idxMass]-_massBinCenters[idxMass-1] - _massStep) > std::numeric_limits<double>::epsilon()) {
```
suhlatwork: this factor 1000 was no GeV to MeV conversion

legordian:
All implemented. Note that merging this pull request will also merge #167, so either merge that one first or only merge this one if #167 is also worthy of inclusion.

suhlatwork:
* file binning does not work as the each file is assigned to an individual bin, so there is one bin containing only real data, one bin with only generated PS, and one bin with only accepted PS, despite the binning maps being the same in these files. A fit cannot be executed.
* `calcIntegrals.py` checks whether all events were used to calculate the amplitudes, this makes usage of the `-n` switch to `calcAmplitudes.py` impossible (even if also `calcIntegrals.py` is called with `-n`. `-n` should probably be removed from both scripts

* in my test case of two event files for two neighbouring mass ranges (1.22 to 1.26 and 1.26 to 1.3) over the full t' range (0.1 to 1.0, divided into eight bins for the integrals) (x3 for real data, generated PS, accepted PS) the integrals for the higher mass bin (1.26 to 1.3) also contain the event metadata of the lower mass bin (1.22 to 1.26). I did not check whether the integrals also include events from this bin. All integral files for the lower mass bin also contain only the metadata of the lower mass bin. This probably also cases warnings (`??? bool rpwa::ampIntegralMatrixMetadata::addKeyFileContent(const string&) [/nfs/freenas/tuph/e18/project/compass/analysis/suhl/rootpwa/rootpwa-devel/decayAmplitude/ampIntegralMatrixMetadata.cc:168]: warning: cannot add keyfile content: already present.`) when running `calcIntegrals.py`. Fitting of the lower mass bin works, fitting of the higher mass bin crashes with a segfault:
```
#6  0x00007f4d626723f9 in SetBranchAddress<double> (ptr=0x0, add=0x423ba98, bname=0x3fac8a8 "mass", this=0x4141c70) at /nfs/mnemosyne/sys/slc6/sw/root/x86-64/6.04.02/root-6.04.02/include/TTree.h:517
#7  rpwa::pwaLikelihood<std::complex<double> >::setOnTheFlyBinning (this=this
entry=0x3c83f80, binningMap=std::map with 2 elements, evtMetas=std::vector of length 2, capacity 2 = {...}) at /nfs/freenas/tuph/e18/project/compass/analysis/suhl/rootpwa/rootpwa-devel/partialWaveFit/pwaLikelihood.cc:1327
#8  0x00007f4d677dc4ca in (anonymous namespace)::pwaLikelihood_setOnTheFlyBinning (self=..., pyBinningMap=..., pyEvtMetas=...) at /nfs/freenas/tuph/e18/project/compass/analysis/suhl/rootpwa/rootpwa-devel/pyInterface/partialWaveFit/pwaLikelihood_py.cc:230
#9  0x00007f4d677dd04b in invoke<boost::python::to_python_value<bool const&>, bool (*)(rpwa::pwaLikelihood<std::complex<double> >&, boost::python::dict, boost::python::list), boost::python::arg_from_python<rpwa::pwaLikelihood<std::complex<double> >&>, boost::python::arg_from_python<boost::python::dict>, boost::python::arg_from_python<boost::python::list> > (ac2=<synthetic pointer>, ac1=<synthetic pointer>, ac0=<synthetic pointer>, f=
0x1598788: 0x7f4d677dbd70 <(anonymous namespace)::pwaLikelihood_setOnTheFlyBinning(rpwa::pwaLikelihood<std::complex<double> >&, boost::python::dict, boost::python::list)>, rc=...) at /nfs/freenas/tuph/e18/project/compass/analysis/suhl/software/boost/boost_1_59_0/include/boost/python/detail/invoke.hpp:75
```

suhlatwork:
Concerning the crash in the likelihood class, the garbage collector of my python version seems to be much more aggressive and deletes the `TFile` from which the event metadata were read, so the event tree is an invalid pointer. A similar thing happens for the amplitudes.

PR #173, opened 2016-06-05
Fix calculation of intensity error

suhlatwork:
This includes #166.

Fix the calculation of the error of an intensity if the full integral matrix was stored. The effect of this bug was somewhat reduced by setting entries corresponding to waves with different reflectivities to zero in the matrix (about a year ago). Now properly fix this also for older fit results.

Also fix a small issue without any effect concerning the `_normIntIndexMap`.


legordian:
Merged, thank you!

PR #174, opened 2016-06-07
Update fit result

suhlatwork:
* no longer store the names of production amplitudes in the fit result, instead store a wave index and a rank for each production amplitude, unfortunately this broke `boost::python` as there were too many arguments to the `fitResult::fill` method, so now do not pass individual vectors, but tuples of vectors
* add a new `spinDensityMatrixTimesAmplitudeMatrix` method to calculate the weights for a subset of waves in one central place, and not have to implement this over and over again
* some cleaning of the code

For my test cases old fit results are still read correctly (after changes to the ROOT I/O rules).

### DIFF
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -240,15 +240,55 @@ fitResult::fill(const unsigned int              nmbEvents,               // numb
 		printWarn << "number of production amplitudes (" << _prodAmps.size() << ") "
 		          << "does not match number of production amplitude names "
 		          << "(" << _prodAmpNames.size() << ")." << endl;
+	if (_waveNames.size() > _prodAmpNames.size())
+		printWarn << "number of wave names (" << _waveNames.size() << ") "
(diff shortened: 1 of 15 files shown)


suhlatwork:
Added some protection for `rootcint` of ROOT 5. We really should start to require ROOT 6!

partialWaveFit/fitResult.cc
```diff
@@ -148,21 +328,31 @@ double
...
 		retval += summands[i];
 	}
 	return retval;
 }
 
+
 std::vector<double>
 fitResult::evidenceComponents() const
 {
 	// make sure evidence can be calculated, i.e.:
```
bgrube: since the returned vector always seems to have 4 elements, wouldn't it make sense to initialize the four values with -infty in the beginning and explicitely set those that are defined?

suhlatwork: I added something like this.

bgrube:
Merged, thank you!

PR #182, opened 2016-07-29
pyInterface: search for amplitude metadata in 'printMetadata'

suhlatwork:
Do not require that 'printMetadata' is called with an argument
specifying which amplitude to print, but loop over the keys of the ROOT
file, and print all amplitude metadata objects that are found. This
covers the typical case of one amplitude per file, but should also be
working if multiple amplitudes are stored in the same file.

### DIFF
--- a/pyInterface/scripts/printMetadata.py
+++ b/pyInterface/scripts/printMetadata.py
@@ -51,7 +51,7 @@ if __name__ == "__main__":
 		# get the amplitude names from the list of keys
 		if key.GetClassName() == "rpwa::amplitudeMetadata":
 			if not key.GetName().endswith(".meta"):
-				pyRootPwa.utils.printErr("key with name '{}' is an amplitude metadata object but does not follow the naming convention.")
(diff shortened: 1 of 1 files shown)


pyInterface/scripts/printMetadata.py
```diff
@@ -17,7 +16,6 @@
 
 	parser = argparse.ArgumentParser(description="print event metadata")
 	parser.add_argument("inputFile", type=str, metavar="inputFile", help="input file in ROOTPWA format")
-	parser.add_argument("-a", "--amplitudeObjectBaseName", type=str, metavar="objectBaseName", help="amplitude object base name (default: try to guess from input file name)")
 	parser.add_argument("-r", action="store_true", dest="recalculateHash", help="recalculate hash and compare it with the stored one (default: %(default)s)")
```
bgrube: Unrelated to your changes: wouldn't
`-v`, `--verify`, `verify hash(es)`
be a better naming for this option?
However, I'm not sure, whether it would be worth breaking the interface.

suhlatwork:
* fixed one string format in an error message added in the previous commit
* renamed the option to verify the hashes, verify actually is a better description for what is done

bgrube:
Merged, thank you.

PR #171, opened 2016-06-03
Add new massDependences and expansions

fkrinner:
Mass dependences:
- Polynomial
- Fourier mode
- Sawtooth
- Arbitrary TFormula
Expansions:
- Taylor
- Fourier
- Chebyshev
- Sawtooth (no extra code was necessary for this, works automatically with the new sawtooth mass-dependence and the old bin-expansion)

PR #183, opened 2016-07-29
Mass dependencies in wave set generator

suhlatwork:
**This is not yet a real pull request!**

To specify the mass dependencies of isobars in the wave set generator, I propose the format indicated in the new version of `userAnalysisWorkspace/3pi.-00/keyfiles/IsobarWhiteList.key`:
- add a new list called `massDependencies` to the `waveSetParameters` group of the configuration file
- this list is a list of groups containing one string value `isobar` and an array of strings `massDeps`
  - the isobar names need to be given as a value of a setting as they cannot be used as names as they contain special characters like brackets
  - multiple mass dependencies might be specified to be able to compare the impact of isobar parametrisations
  - the same translation from the name to the mass dependency as in the wave description is used, so the empty string is short for `relativisticBreitWigner`
- do we need to be able to handle special mass dependencies like `binned` here? this would make the task a lot more complex as then additional arguments would need to be specified and handled in the wave set generator; we would then also need a way to create the keyfile expansion or move the keyfile expansion from the wave description into the wave set generator.


bgrube:
I would propose to approach this slightly differently.

`generateWaveSet` already has the `-d` option, where the user can provide a file with a list of allowed decay modes for the isobars. It was necessary for the 5-pion analysis (see  [userAnalysisWorkspace/5pi.-+-+-/keyfiles/5piParticleDecays.txt](userAnalysisWorkspace/5pi.-+-+-/keyfiles/5piParticleDecays.txt)). Up to know it was not used for the 3pi channels as all isobars decay into 2 pions. But I think this could be a better place to store the information about mass parametrizations. One would just need to extend the entries in the `decays` list, which already define the decay products and optionally the orbital angular momentum. So, one could write e.g.
```
particleDecayList = ({

		name = "sigma0";
		decays = (
			{
				products = ["pi-", "pi+"];
				massDeps = [ "piPiSWaveAuMorganPenningtonKachaev" ];
			}
		);

		name = "f0(980)0";
		decays = (
			{
				products = ["pi-", "pi+"];
				massDeps = [ "", "f_0(980)", "f_0(980)Flatte" ];
			}
		);

	})
```
This would require an additional loop over the mass dependences in `waveSetGenerator`. The advantage of this approach is that one has full freedom to specify different mass dependences for different decay modes. This is, of course, also a drawback, since if one wants to use the same parametrization for several decay modes one has to repeat it several times.

suhlatwork:
Your proposal turned out to be much simpler to implement than what I had in mind (at least the reading of the config file, the setting of the mass dependencies is the same).

My current proposal still should not be merged. Different parametrisations of the same isobar will lead to the same wave name, and 'generateWaveSet' will complain about this. I will wait for #171 as this already includes a patch making the required changes a bit more simple.

Still to be done:
- [ ] Unique naming of waves with different mass dependencies
- [ ] Add back different parametrisations for the f0(980) in the 3pi scripts
- [ ] Adapt `testMC.sh`

suhlatwork:
This pull request should now be ready for its final review. Things to discuss:
* I could not think of a better name for the function to get the string used in the wave name than `parentForWaveName`
* the names of mass dependencies are quite long at the moment

suhlatwork:
Reduced the number of hard-coded mass-dependence names to basically zero outside the corresponding classes. I am not entirely sure this actually improved the code.

decayAmplitude/waveSetGenerator.cc
```diff
@@ -565,6 +584,62 @@ waveSetGenerator::createNewDecayTopology(const isobarDecayTopology&  parentDecay
 }
 
 
+vector<isobarDecayTopologyPtr>
+waveSetGenerator::setMassDepsForDecayTopology(const isobarDecayTopologyPtr         decayTopology,
+                                              const particleProperties&            isobar,
+                                              const particleProperties::decayMode& decayMode)
+{
+	// set mass dependencies of isobar decay
+	set<string> usedMassDependencies;
```
bgrube: maybe rename to usedMassDeps to be consistent

decayAmplitude/waveSetGenerator.cc
```diff
@@ -565,6 +584,62 @@ waveSetGenerator::createNewDecayTopology(const isobarDecayTopology&  parentDecay
...
+	//    (there might be multiple matches (e.g. a special mass depedence for a
+	//    particular L))
+	for (size_t iDecay = 0; iDecay < isobar.nmbDecays(); ++iDecay) {
+		if (not (isobar.decayModes()[iDecay] == decayMode))
+			continue;
+
+		//  2. loop over mass dependencies
+		const vector<string>& massDeps = isobar.decayModes()[iDecay]._massDependencies;
+		if (massDeps.size() != 0) {
+			for (vector<string>::const_iterator massDepType = massDeps.begin(); massDepType != massDeps.end(); ++massDepType) {
```
bgrube: maybe rename massDepType to massDep

decayAmplitude/waveSetGenerator.cc
```diff
@@ -565,6 +584,62 @@ waveSetGenerator::createNewDecayTopology(const isobarDecayTopology&  parentDecay
...
+				if (usedMassDependencies.count(*massDepType) != 0)
+					continue;
+				usedMassDependencies.insert(*massDepType);
+				if (*massDepType == relativisticBreitWigner::Name()) usedMassDependencies.insert("");
+				if (*massDepType == "") usedMassDependencies.insert(relativisticBreitWigner::Name());
+
+				massDependencePtr massDep = createMassDependence(*massDepType);
```
bgrube: variable not needed anywhere else; maybe move createMassDependence() directly into if statement below

suhlatwork:
Updated the pull request.

The mass dependence still needs to be specified for each set of decay products. While it is for sure possible to allow setting a mass dependence that would be valid for all decays, this is probably not worth the time to implement this.

bgrube:
Agreed, this would not be worth the effort.

bgrube:
Merged. Thank you!

PR #184, opened 2016-08-01
Rename `genpw` to `generatePhaseSpace`

suhlatwork:
Rename the Python script `genpw` to `generatePhaseSpace` removing the previous C++ executable. It seems to be required to remove the content of `build/bin/` for this change to take effect (`rm -rf build/bin/*`), otherwise the link from this directory to the script will not be created. This changes the options with which `generatePhaseSpace` can be called in one case, simply removing `-r` (without removing the argument to `-r` should work).


bgrube:
Merged, thank you.

PR #185, opened 2016-08-05
Check pylint version

suhlatwork:
Require at least version 1.5.5 of pylint and only add the tests if at least this version was found. Older versions might be fine but I cannot test starting from which version they work.


suhlatwork:
@stefan-wallner

bgrube:
Merged, thank you.

PR #186, opened 2016-08-08
Travis update

suhlatwork:
Still work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,13 @@
 sudo: required
 dist: trusty
 
+# test with two different settings for the optional packages
+env:
+    # without any optional package
+    - TRAVIS_RPWA_USE_NLOPT=0
+    # with all optional packages
+    - TRAVIS_RPWA_USE_NLOPT=1
+
 language: cpp
 
 # send mails if a build fails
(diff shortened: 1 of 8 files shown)


suhlatwork:
done:
* Travis is now using Ubuntu 14.04, this includes a more recent version of GCC (now 4.8.4), Python (now 2.7.10). a hackish workaround for Python (otherwise the build would fail with `relocation R_X86_64_32S against `_Py_NotImplementedStruct' can not be used when making a shared object` when linking the Python bindings) was required
* use libconfig, yaml-cpp and NLopt provided by Ubuntu so we can save some time during the build, the provided versions are recent enough
* update the versions of CMake, BOOST and ROOT to the latest versions, for ROOT this means we are finally using ROOT6. ROOT is now installed from the prebuilt package from root.cern.ch
* we now have two builds, one without any optional dependency, the other with NLopt and BAT. The latter should probably be extended to also include the optional MPI and Cuda stuff.
* pylint and the `testMC.sh` script are executed during the build

to discuss:
* for the moment BOOST is installed from the individual modules, this is quite tedious work to get all the dependencies, it might be safer to go back to install everything even though we only need about half
* the currently last commit should not be merged, that means waiting for #187, #188; also a better way to exclude `pwaNloptFit` from `testMC.sh` is required
* pylint is fixed to version 1.5.5 for the moment, more recent versions are available, but not requiring a fixed version might cause build failures if a new check as added in pylint

.travis.yml
```diff
@@ -1,31 +1,53 @@
...
-cache: apt
-notifications:
-    email:
-        recipients: rootpwa-push-notifications@lists.sourceforge.net
-        on_success: change
-        on_failure: always
+    # run 'make'
+    # - if 'make' was not successful, fail the build
+    # - if 'make' was successful, run 'make test'
+    # - if 'make test' was not successful, print the log and fail the build
```
bgrube: add
# - if 'make test' was successful, run testMC.sh

travisCI/install-boost.sh
```diff
@@ -1,4 +1,30 @@
...
 set -exv
-wget 'http://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.gz'
-tar -xzf boost_1_58_0.tar.gz
+
+LIBS="accumulators algorithm align any array assert assign atomic bimap bind chrono concept_check config container conversion core date_time detail dynamic_bitset exception foreach function function_types functional fusion graph graph_parallel integer intrusive io iterator lexical_cast math move mpi mpl multi_array multi_index optional parameter predef preprocessor property_map proto python range ratio rational regex serialization smart_ptr spirit static_assert system test thread throw_exception timer tokenizer tti tuple type_index type_traits typeof unordered utility xpressive"
+TOOLS="build inspect"
+EXTRA="interval numeric_conversion ublas"
+
+# download required files
+wget https://github.com/boostorg/boost/archive/boost-1.61.0.tar.gz -O boost.tar.gz
```
bgrube: It would be nice to set the boost version via a variable. This would make switching to different versions easier.

.travis.yml
suhlatwork: I added this line; but have no idea why github does not understand this diff is outdated

bgrube:
Merged, thank you.

PR #187, opened 2016-08-08
buildSystem: fix NLopt header when using system library

suhlatwork:
Fix the name of the header file searched when the system library should
be used (i.e., the NLOPT environment variable is not set). The name jad
the wrong capitalization and we are actually interested in the header
file for C++, not pure C.


bgrube:
Merged, thank you.

PR #188, opened 2016-08-08
buildSystem: 'pwaNloptFit' without NLopt library

suhlatwork:
Do not install the 'pwaNloptFit' Python script and do not add a test for this script if ROOTPWA is compiled without support for NLopt.


bgrube:
Merged, thank you.

PR #189, opened 2016-08-10
pyInterface: fix for new pylint versions

suhlatwork:
Directly loop over dictionaries instead of instead of calling 'keys()'. This fixes pylint check C0201 (consider-iterating-dictionary).


bgrube:
Merged, thank you.

PR #190, opened 2016-10-06
_integralsOnTheFly.py: Small bug fix

fkrinner:
The `continue` statement only affected the loop over kinematic variables, not the event loop


legordian:
Merged, thank you!

PR #191, opened 2016-10-17
buildSystem: Fix compilation errors

bgrube:
Fix compilation errors with gcc 6.2.0 (Ubuntu 16.10).

### DIFF
--- a/pyInterface/CMakeLists.txt
+++ b/pyInterface/CMakeLists.txt
@@ -160,12 +160,28 @@ add_custom_target(
 # pylint tests
 if(PYLINT_FOUND)
 	# check 'pyRootPwa' module
-	add_test(pyLintPyRootPwa ${PYLINT_EXECUTABLE} --reports=n --rcfile=${CMAKE_CURRENT_SOURCE_DIR}/pylintrc pyRootPwa)
+	add_test(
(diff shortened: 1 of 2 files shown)


decayAmplitude/decayTopology.cc
```diff
@@ -145,9 +145,10 @@ decayTopology::doClone(const bool cloneFsParticles,
 		// clone vertex
 		if (   isDecayVertex(vert)
 		    or isProductionVertex(vert)
-		    or (cloneFsParticles and isFsVertex(vert)))
+		    or (cloneFsParticles and isFsVertex(vert))) {
 			topoClone->isDecayVertex(vert);
-			newVerts.push_back(topoClone->cloneNode(*iNd));
+		}
+		newVerts.push_back(topoClone->cloneNode(*iNd));
```
suhlatwork: Line 149 does not have any effect, so with your current proposal the complete ```if``` statement including its action (lines 146 to 150 with the patch) could be removed. Are you sure this is correct? Probably the indentation checker really found a bug here.

bgrube: Indeed this is a bug. It is fixed in 5c2c49e.

PR #193, opened 2016-11-09
Require 'numpy'

suhlatwork:
And a Cmake module to search for 'numpy'. 'numpy' is required for a couple of Python scripts, so it should be a hard requirement to be able to build ROOTPWA.


bgrube:
Merged, thank you.

PR #194, opened 2016-11-16
partialWaveFit: compatibility with ROOT 6.08

cdreisbach:
TObject.h does not include TBuffer.h anymore.
(see https://root.cern.ch/doc/v608/release-notes.html#core-libraries)

Added include of TBuffer.h in complexMatrix.cc


suhlatwork:
Merged. Thank you!

PR #195, opened 2016-11-18
generatePhaseSpace: Error handling if no output file

cdreisbach:
If a not-existing output file is passed as argument to
generatePhaseSpace, the usage of this file as data
file via ROOT.Open(filename) gives an error

Added a check whether the args.outputFileName is a valid
file


suhlatwork:
This will stop `generatePhaseSpace` from working at all, won't it? With your patch `generatePhaseSpace` stops working with non-existing output files, while an already existing output file will cause a stop when trying to create the output ROOT file.

cdreisbach:
Yes, your are right. Thank you.

PR #197, opened 2016-12-05
Set git hash in ampIntegralMatrixMetadata

suhlatwork:
Set the hash to a value if all amplitude files have the same hash, and this hash is the same as the current ROOTPWA hash. Otherwise leave it empty (as is currently the case).

### DIFF
--- a/pyInterface/module/_integrals.py
+++ b/pyInterface/module/_integrals.py
@@ -8,6 +8,7 @@ def calcIntegrals(integralFileName, eventAndAmpFileDict, multiBin, weightFileNam
 		pyRootPwa.utils.printWarn("cannot open output file '" + integralFileName + "'. Aborting...")
 		return False
 	integralMetaData = pyRootPwa.core.ampIntegralMatrixMetadata()
+	integralMetaData.setGitHash(pyRootPwa.core.gitHash())
 	integralMetaData.setBinningMap(multiBin.boundaries)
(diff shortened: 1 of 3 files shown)


suhlatwork:
Now always set the git hash to the revision of the `calcIntegral` executable.

pyInterface/module/_integrals.py
```diff
@@ -8,8 +8,12 @@ def calcIntegrals(integralFileName, eventAndAmpFileDict, multiBin, weightFileNam
 		pyRootPwa.utils.printWarn("cannot open output file '" + integralFileName + "'. Aborting...")
 		return False
 	integralMetaData = pyRootPwa.core.ampIntegralMatrixMetadata()
+	integralMetaData.setGitHash(pyRootPwa.core.gitHash())
 	integralMetaData.setBinningMap(multiBin.boundaries)
 	integrals = []
+	# initialize the git hash with the hash of the current ROOTPWA. this
+	# is compared to the ones from the amplitude files and only kept if all
+	# agree. otherwise an empty string will be set.
 	for eventFileName, ampFileDict in eventAndAmpFileDict.iteritems():
```
bgrube: The above comments need to be removes

suhlatwork:
Removed the outdated comment, thank you for noticing.

bgrube:
Merged. Thank you!

PR #199, opened 2016-12-05
Normalization of fit result

suhlatwork:
`fitResult::_normNmbEvents` is not taken into account while creating information like intensity, ... This causes wrong results in case the partial-wave fit is performed without normalization of the amplitudes in ROOTPWA, or when naively importing fit results from COMPASSPWA (one can apply the normalization with the number of events to the production amplitudes during the conversion, of course).

It needs to be decided whether only intensities should be normalized with the number of events, or whether already the spin-density matrix elements need to have this.


suhlatwork:
Two different cases have to be considered:

1. The transition amplitudes are not normalized to the number of events, but the integrals are normalized such that the diagonal entries of the phase-space integral matrix are 1. This is the case for fit results imported from COMPASSPWA.
To make the output of the 'fitResult' class directly comparable with the plots from COMPASSPWA, '_normNmbEvents' is multiplied on the spin-density matrix elements.

2. For the ROOTPWA fits without normalized amplitudes in addition the integral matrices are not normalized. This is taken into account for the intensities, but not for the spin-density elements (and I do not think that this is actually possible).

bgrube:
Merged. Thank you!

PR #200, opened 2016-12-07
Another fix for ROOT 6.08

suhlatwork:
Finally understood the second issue with ROOT 6.08 observed by @fkaspar and @cdreisbach. If also BAT is to be used, then an additional declaration of `TClonesArray` is required to compile the event file writer.


bgrube:
Merged. Thanks!

PR #201, opened 2016-12-07
buildSystem: correct library name in error message

suhlatwork:
One error message while searching for BAT in CMake still printed
'libconfig' (the template CMake module used for the BAT one). Corrected
this.


bgrube:
Merged, thank you.

PR #202, opened 2016-12-08
Require C++11

suhlatwork:
Require C++11 while configuring ROOTPWA. Do not yet actually modify any code (apart from removing non-C++11 workarounds).


bgrube:
Merged, thank you.

PR #203, opened 2016-12-16
partialWaveFit: less warnings for fit result without matrices

suhlatwork:
Reduce the number of warnings printed while filling a 'rpwa::fitResult'
object with empty covariance, normalization or acceptance integrals. In
that case only warn once (about an empty matrix) instead of (in case of
the covariance matrix) complain for each production amplitude.


bgrube:
Merged, thank you.

PR #204, opened 2016-12-16
Printing

suhlatwork:
Reduce the number of messages printed during a partial-wave fit (and in other cases as well, but that is the case were they are really annoying as they fill up all the logfiles with irrelevant information).

### DIFF
--- a/partialWaveFit/pwaLikelihood.cc
+++ b/partialWaveFit/pwaLikelihood.cc
@@ -1320,6 +1320,11 @@ pwaLikelihood<complexT>::finishInit()
 			 _decayAmps.num_elements(), _nmbEvents, _nmbWavesRefl, true);
 #endif
 
+	printSucc << "set up likelihood function for rank-" << _rank << " fit with "
+	          << _nmbWaves << " wave" << ((_nmbWaves != 1) ? "s" : "") << " (excluding 'flat' wave; "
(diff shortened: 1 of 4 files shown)


partialWaveFit/pwaLikelihood.cc
```diff
@@ -1414,6 +1414,9 @@ pwaLikelihood<complexT>::readWaveList(const vector<waveDescThresType>& waveDescT
 			_waveAmpAdded  [iRefl][iWave] = false;
 			_waveParams.insert(pair<string, pair<unsigned int, unsigned int> >(waveNames[iRefl][iWave], pair<unsigned int, unsigned int>(iRefl, iWave)));
 		}
+
+	printInfo << "set up likelihood function for " << _nmbWaves << " wave" << ((_nmbWaves != 1) ? "s" : "") << " (excluding 'flat' wave)." << endl;
+
```
bgrube: This is nice. Maybe one could add the rank and the number of positive- and negative-reflectvity waves to the printout.

suhlatwork:
Previously (excluding this pull request) the messages concerning the creation of the decay topologies were printed from the the `pwaLikelihood::readWaveList` function. So in the first version of this pull request also the message concerning how many waves are used was added in this function.

Now that also the rank should be printed, this message has been moved to `pwaLikelihood::finishInit`. It is the success message summarizing the main properties of the likelihood object.

bgrube:
Merged. Thank you!

PR #205, opened 2017-01-31
buildSystem: ask for Boost libraries with multithreading support

suhlatwork:
Set to 'Boost_USE_MULTITHREADED' variable to 'ON' when searching for the
Boost libraries. This changes the long file/library names searched for
when running Cmake, it does not change the name as created if Boost is
compiled as explained in 'README.md'. The long name also includes the
used compiler and various other information. The default of Boost seems
to be to use multithreading since a long time, so all installations
using the long names when creating the libraries require this change.

 Enabling support for mulitthreading seems to be the
default of Boost since long time. T


bgrube:
Merged, thank you.

PR #206, opened 2017-02-01
nBodyPhaseSpace: random numbers must be strictly increasing

suhlatwork:
The random numbers used to calculate the intermediate masses have to be
strictly increasing as there is a strict requirement on the masses of
the states (see eq. 20 in F. James "Monte Carlo Phase Space"). Repeat
the creation of the list of random numbers until this is the case.


bgrube:
Merged. Thanks!

PR #207, opened 2017-02-16
particleData: update to PDG 2016 values

suhlatwork:
Update the particle data table to the values from the 2016 edition of
the PDG.

### DIFF
--- a/particleData/particleDataTable2016.txt
+++ b/particleData/particleDataTable2016.txt
@@ -1,0 +1,321 @@
+##########################################################################
+#
+#    Copyright 2012,2017
+#
+#    This file is part of rootpwa
+#
+#    rootpwa is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


bgrube:
Thanks a lot Sebastian! However, I would propose to keep a copy of the old file (e.g. as "particleDataTable2012.txt"). This way, we have as default the latest and greatest and at the same time it is easy to repeat fits from the publications.

suhlatwork:
Keep the old file around.

stefan-wallner:
I would not update the `particleData/particleDataTable.txt` file, because everyone who updates his ROOTPWA version without carefully checking the commits can obtain (slightly) different results using the updated table without knowing the reason. 
I would create a `particleData/particleDataTable2016.txt` file and perhaps even move `particleData/particleDataTable.txt` to `particleData/particleDataTable2012.txt`. 
If one wants to have the newest one as default, one may adjust `rootpwa.config` accordingly.

suhlatwork:
Add the 2016 data as a new file. Added a card to Trello to take care of making this the default.

bgrube:
Merged. Thank you!

PR #208, opened 2017-04-27
Remove warning when using more amplitude files for calculating the integrals

stefan-wallner:
When using more amplitude files for calculating the integrals, the
keyfile content was added n-files times, which created a warning. 
Now, the keyfile content is added only once and checked for any further
amplitude file.

### DIFF
--- a/pyInterface/module/_integrals.py
+++ b/pyInterface/module/_integrals.py
@@ -38,7 +38,7 @@ def calcIntegrals(integralFileName, eventAndAmpFileDict, multiBin, weightFileNam
 				if not integralMetaData.hasKeyFileContent(ampMeta.keyfileContent()):
 					pyRootPwa.utils.printErr("keyfile content of additional eventFiledID missing in first eventFieldID.")
 					return False
-					
+
 			if not integralMetaData.addAmplitudeHash(ampMeta.contentHash()):
 				# This error is not fatal, since in special cases the same hash can appear twice:
 				# e.g. in freed-isobar analyses with spin zero, the angular dependences are constant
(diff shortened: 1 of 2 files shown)


pyInterface/module/_integrals.py
```diff
@@ -30,7 +30,14 @@ def calcIntegrals(integralFileName, eventAndAmpFileDict, multiBin, weightFileNam
 				pyRootPwa.utils.printErr("could not read metadata from amplitude file '" + ampFileName + "'. Aborting...")
 				return False
 			ampMetas.append(ampMeta)
-			integralMetaData.addKeyFileContent(ampMeta.keyfileContent())
+
+			if len(integrals) == 1: # first eventFieldID of the current bin
+				if not integralMetaData.addKeyFileContent(ampMeta.keyfileContent()):
+					pyRootPwa.utils.printWarn("could not add the keyfile content.")
```
suhlatwork: `addKeyFileContent` already prints a similar error message, so I think this could be skipped here.
Instead a `return false` should be added to signal this error.

pyInterface/module/_integrals.py
```diff
@@ -30,7 +30,14 @@ def calcIntegrals(integralFileName, eventAndAmpFileDict, multiBin, weightFileNam
...
+			else:
+				if not integralMetaData.hasKeyFileContent(ampMeta.keyfileContent()):
+					pyRootPwa.utils.printWarn("keyfile content of additional eventFiledID missing in first eventFieldID.")
```
suhlatwork: Again, I think that this should be a fatal error, so probably `printErr` instead of `printWarn` should be used, also here `return false` should be added.

suhlatwork:
Merged with fixing one whitespace error. Thank you!

PR #209, opened 2017-04-28
Correction version of pyRootPwa in pyLintPyRootPwa test

stefan-wallner:
The pyLintPyRootPwa test used the pyRootPwa version from PYTHONPATH,
which not necessarily is the current build folder.
This is now fixed by setting the PYTHONPATH environment variable
in advance.


suhlatwork:
Merged with hopefully not creating too much of a mess in a1bab19e04f88244e1414ad2b3215851bb0b3a8c.

PR #210, opened 2017-04-28
Binningmap in fitresult

stefan-wallner:


highLevelInterface/pwaFit.h
```diff
@@ -9,8 +9,7 @@ namespace rpwa {
 	namespace hli {
 
 		rpwa::fitResultPtr pwaFit(const rpwa::pwaLikelihood<std::complex<double> >& L,
-		                          const double                                      massBinMin = 0.,
-		                          const double                                      massBinMax = 0.,
+		                          const binningMapType& binningMap,
```
suhlatwork: Please align the `binningMap` with the other variable names.

partialWaveFit/addIntegralsToFitResult.cc
```diff
@@ -187,7 +187,7 @@ main(int    argc,
 		inResultTree->GetEntry(i);
 		const unsigned int             nmbEvents              = inResult->nmbEvents();
 		const unsigned int             normNmbEvents          = inResult->normNmbEvents();
-		const double                   massBinCenter          = inResult->massBinCenter();
+		const binningMapType           binningMap             = inResult->binningMap();
```
suhlatwork: this could be a reference

partialWaveFit/fitResult.cc
```diff
@@ -343,6 +342,14 @@ fitResult::fill(const fitResult& result)
...
 }
 
+multibinCenterType fitResult::multibinCenter() const
+{
+	multibinCenterType multibinCenter;
+	for (binningMapType::const_iterator it = _binningMap.begin(); it != _binningMap.end(); ++it) {
+		multibinCenter[it->first] = 0.5 * (it->second.first + it->second.second);
+	}
+	return multibinCenter;
+}
```
suhlatwork: Does the mean value of the limits make sense for anything but the mass? For example, for tPrime the mean of the individual values and the mean of the limits will typically systematically deviate.

suhlatwork: functions should be surrounded by two empty lines, the return type of a function should be on a separate line.

partialWaveFit/fitResult.h
```diff
@@ -132,18 +133,22 @@ namespace rpwa {
...
-		bool                hasHessian        () const { return _hasHessian;        }  ///< returns whether Hessian matrix has been calculated successfully
-		unsigned int        nmbWaves          () const { return _waveNames.size();  }  ///< returns number of waves in fit
-		unsigned int        nmbProdAmps       () const { return _prodAmps.size();   }  ///< returns number of production amplitudes
+		void                  setBinningMap     (const binningMapType& binningMap)   { _binningMap = binningMap; } ///< set binning map
+
+		unsigned int          nmbEvents         () const { return _nmbEvents;               }  ///< returns number of events in bin
+		unsigned int          normNmbEvents     () const { return _normNmbEvents;           }  ///< returns number of events to normalize to
+		double                massBinCenter     () const { return multibinCenter()["mass"]; }  ///< returns center value of mass bin
+		multibinCenterType    multibinCenter    () const;                                      ///< returns center value of bin in each dimension
+		const binningMapType& binningMap        () const {return _binningMap;               }  ///< returns binning map
```
suhlatwork: Please align the `return` with the other returns.

partialWaveFit/fitResult.h
```diff
@@ -132,18 +133,22 @@ namespace rpwa {
...
-		std::vector<double> evidenceComponents() const;                                ///< returns a vector { maxLogL, ln(sqrt((2\pi)^m*cov)), -ln(V_A^k), \sum(S_\alpha) }, i.e. evidence = sum(evidenceComponents[i]) for i in {1,2,3,4}
-		unsigned int        rank              () const { return _rank;              }  ///< returns rank of fit
-		bool                covMatrixValid    () const { return _covMatrixValid;    }
-		bool                converged         () const { return _converged;         }  ///< returns whether fit has converged (according to minimizer)
-		bool                hasHessian        () const { return _hasHessian;        }  ///< returns whether Hessian matrix has been calculated successfully
-		unsigned int        nmbWaves          () const { return _waveNames.size();  }  ///< returns number of waves in fit
-		unsigned int        nmbProdAmps       () const { return _prodAmps.size();   }  ///< returns number of production amplitudes
+		void                  setBinningMap     (const binningMapType& binningMap)   { _binningMap = binningMap; } ///< set binning map
```
suhlatwork: Is it really required to set a new binning map when not filling a new fit result?

partialWaveFit/fitResult.h
```diff
@@ -132,18 +133,22 @@ namespace rpwa {
...
+
+		unsigned int          nmbEvents         () const { return _nmbEvents;               }  ///< returns number of events in bin
+		unsigned int          normNmbEvents     () const { return _normNmbEvents;           }  ///< returns number of events to normalize to
+		double                massBinCenter     () const { return multibinCenter()["mass"]; }  ///< returns center value of mass bin
```
suhlatwork: I am not sure if we should not entirely remove this function. This would probably require quite some changes in code using the fit result, but I am not sure the binning maps are even required to contain a `mass` variable by the file manager.

partialWaveFit/linkdef.h
```diff
@@ -73,5 +73,19 @@
 }                                                                                                   \
               }";
 
+#pragma read                                                                                        \
+        sourceClass="rpwa::fitResult"                                                               \
+        source="Double_t _massBinCenter"                                                            \
+        version="[-6]"                                                                              \
+        targetClass="rpwa::fitResult"                                                               \
+        target=""                                                                        \
```
suhlatwork: align the trailing `\` with the others.

partialWaveFit/linkdef.h
```diff
@@ -73,5 +73,19 @@
...
+        include="utility"                                                                           \
```
suhlatwork: do you need the `utility` include if you do not use `std::make_pair`?

partialWaveFit/fitResult.h
suhlatwork: Given that this should not be used outside the scheme evolution, can we somehow hide this?

pyInterface/bindings/highLevelInterface/pwaFit_py.cc
```diff
@@ -8,16 +8,36 @@
...
 
+namespace {
+
+rpwa::fitResultPtr pwaFit_pwaFit(const rpwa::pwaLikelihood<std::complex<double> >& L,
+									const bp::dict&                                   binningMap,
+									const unsigned int                                seed = 0,
+									const std::string&                                startValFileName = "",
+									const bool                                        checkHessian = false,
+									const bool                                        saveSpace = false,
+									const bool                                        verbose = false)
```
suhlatwork: Please indent by one tab inside the namespace and align the arguments with spaces instead of tabs.

highLevelInterface/pwaFit.cc
```diff
@@ -63,8 +63,7 @@ namespace ROOT {
 
 fitResultPtr
 rpwa::hli::pwaFit(const pwaLikelihood<complex<double> >& L,
-                  const double                           massBinMin,
-                  const double                           massBinMax,
+                  const binningMapType&                  binningMap,
```
suhlatwork: This could take an empty map as a default argument, you in any case do this for the Python interface.

pyInterface/bindings/partialWaveFit/fitResult_py.cc
```diff
@@ -16,7 +16,7 @@ namespace {
 	void fitResult_fill_1(rpwa::fitResult& self,
 	                      const unsigned int                        nmbEvents,
 	                      const unsigned int                        normNmbEvents,
-	                      const double                              massBinCenter,
+	                      const bp::dict& binningMap,
```
suhlatwork: please align `binningMap` with the other arguments.

pyInterface/bindings/partialWaveFit/fitResult_py.cc
```diff
@@ -60,14 +60,53 @@ namespace {
...
+		std::map<std::string, std::pair<double, double> > cBinningMap;
+		const bp::list keys = binningMap.keys();
+		for (int i = 0; i < bp::len(keys); ++i) {
+			std::string binningVar = bp::extract<std::string>(keys[i]);
+			double lowerBound      = bp::extract<double>(binningMap[binningVar][0]);
+			double upperBound      = bp::extract<double>(binningMap[binningVar][1]);
+			cBinningMap[binningVar] = std::pair<double, double>(lowerBound, upperBound);
+		}
+		self.setBinningMap(cBinningMap);
+	}
```
suhlatwork: given that the `setBinningMap` function should only be called by the ROOT streamer, this function should probably not be exported to Python.

pyInterface/module/_fit.py
```diff
@@ -127,8 +120,7 @@ def pwaNloptFit(eventAndAmpFileDict,
 	fitResults = [ ]
 	for fitSeed in seeds:
 		fitResult = pyRootPwa.core.pwaNloptFit(likelihood       = likelihood,
-		                                       massBinMin       = lowerBound,
-		                                       massBinMax       = upperBound,
+											   binningMap       = multiBin.boundaries,
```
suhlatwork: indentation

highLevelInterface/pwaFit.h
suhlatwork: Actually the default argument should probably go here.

partialWaveFit/fitResult.cc
stefan-wallner: I just wanted to implement a general function that calculates the center of the multibin. 
At the moment, is is anyway only used for the mass, so it could be removed, but if we later have another binning variable (for what ever reason), a general bin-center function could be useful. 
What do you think? @suhlatwork

partialWaveFit/fitResult.h
stefan-wallner: As far as I can see, is is not required by the file manager. 
I think, avoiding refactoring quite some amount of the code, we should require it!

highLevelInterface/pwaFit.cc
```diff
@@ -97,10 +96,18 @@ rpwa::hli::pwaFit(const pwaLikelihood<complex<double> >& L,
...
 
+	const double massBinMin    = binningMap.at("mass").first;
+	const double massBinMax    = binningMap.at("mass").second;
+
 	// report parameters
 	printInfo << "running pwaFit with the following parameters:" << endl;
-	cout << "    mass bin [" << massBinMin << ", " << massBinMax << "] GeV/c^2" << endl
-	     << "    seed for random start values ................... "  << seed                    << endl
+	cout << "    mass bin [" << massBinMin << ", " << massBinMax << "] GeV/c^2" << endl;
+	for(const auto& variable_borders: binningMap){
```
bgrube: maybe one could improve the naming: why not just call variable_borders "bin", or "binBorders", at least "variableBorders".

storageFormats/eventMetadata.h
```diff
@@ -11,6 +11,8 @@ class TTree;
 
 
 namespace rpwa {
+	typedef std::map<std::string, std::pair<double, double> > binningMapType;
+	typedef std::map<std::string, double > multibinCenterType;
```
bgrube: Maybe "binCenterType" would be a better name for "multibinCenterType", beause it just holds variable name and bin center. Not sure we should go that far, but in principle "binRangeType" would be a better name for "binningMapType". However this would probably require changes in a lot of places.

stefan-wallner: I agree with your name suggestions.
We should only agree on a common name for a multi-dimensional bin. Should it be called 'bin', 'multibin', or 'cell'?
In the python interface, it is called multibin, e.g. the corresponding class is called  'multiBin'.
The changes are no problem using Eclips code refactoring tools.

stefan-wallner: After our discussion, my suggestion would be #223

suhlatwork:
Merged as commit 3b73d9282a3da77a8ea5a21b81f4ac53fa547245 with mostly cosmetic changes:
* whitespaces and indentations in output on the console
* in the Python interface variables that need to be translated should be called `pyVar` for the Python thing and `var` for the C++ thing

Thank you!

PR #211, opened 2017-05-31
Implement rounding to 14 significant digits for on-the-fly binning

stefan-wallner:
When on-the-fly bins are generated from a range and a number of bins, now the bin borders are rounded to 14 significant digites to get rid of the nummeric effects of the boundary calculation. (e.g. 2.4000000000000004)

### DIFF
--- a/pyInterface/module/_config.py
+++ b/pyInterface/module/_config.py
@@ -207,7 +207,7 @@ def _binningInputHandlingForCaseTwoAndThree(inputVal):
 			nBins = int(axesPartition[2])
 			for i in xrange(nBins+1):
 				# round to 14 significant digits to get rid of numeric artifacts of the bin borders
-				expandedBoundaryList.append(_roundToNSignificantDigits( xMin + (xMax-xMin)/nBins * i, 14))
+				expandedBoundaryList.append(_roundToNSignificantDigits(xMin + (xMax-xMin)/nBins * i, 14))
 			inputVal[key] = expandedBoundaryList
 	return inputVal
 
--- a/pyInterface/module/_config.py
+++ b/pyInterface/module/_config.py
@@ -1,4 +1,4 @@
-
+import math
 import ast
 import ConfigParser
 import os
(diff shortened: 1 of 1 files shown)


pyInterface/module/_config.py
```diff
@@ -171,6 +171,11 @@ def _readBinning(binningString):
 				return []
 	return retval
```
suhlatwork: there should be two empty lines to separate functions

pyInterface/module/_config.py
```diff
@@ -171,6 +171,11 @@ def _readBinning(binningString):
...
 
+def _roundSignificantDigits(value, significantDigits):
+	oomValue = int( math.floor(math.log10(abs(value))))
```
suhlatwork: there is no space to balance the one after `int(`, please write either `int( math.floor(math.log10(abs(value))) )` or `int(math.floor(math.log10(abs(value))))`

pyInterface/module/_config.py
```diff
@@ -171,6 +171,11 @@ def _readBinning(binningString):
...
+	nDigits = significantDigits - 1 - oomValue
+	return round(value,nDigits)
```
suhlatwork: arguments should be separated by `, `. please add the space

pyInterface/module/_config.py
```diff
@@ -196,8 +201,12 @@ def _binningInputHandlingForCaseTwoAndThree(inputVal):
...
 				raise ValueError(errMsg)
 			expandedBoundaryList = []
-			for i in xrange(axesPartition[2]+1):
-				expandedBoundaryList.append(float(axesPartition[0]) + (float(axesPartition[1] - axesPartition[0]) / float(axesPartition[2])) * i)
+			xMin = float(axesPartition[0])
+			xMax = float(axesPartition[1])
+			nBins = int(axesPartition[2])
+			for i in xrange(nBins+1):
+				# round to 14 significant digits to get rid of numeric artifacts of the bin borders
+				expandedBoundaryList.append( _roundSignificantDigits( xMin + (xMax-xMin)/float(nBins) * i, 14) )
```
suhlatwork: please remove the space after `_roundSignificantDigits(`

pyInterface/module/_config.py
```diff
@@ -171,6 +171,11 @@ def _readBinning(binningString):
 				return []
 	return retval
 
+def _roundSignificantDigits(value, significantDigits):
```
suhlatwork: IMHO `_roundToSignificantDigits`/`_roundToNSignificantDigits` would be a better name

pyInterface/module/_config.py
suhlatwork: the `float` cast of `nBins` should not change anything here, right?

stefan-wallner:
Thanks for your comments. I agree with all of them.

suhlatwork:
Merged, thank you!

PR #213, opened 2017-06-02
storageFormats: Add variable to store auxiliary values in eventMetadata

stefan-wallner:


storageFormats/eventMetadata.h
```diff
@@ -40,6 +40,10 @@ namespace rpwa {
 		const std::vector<std::string>& productionKinematicsParticleNames() const { return _productionKinematicsParticleNames; }
 		const std::vector<std::string>& decayKinematicsParticleNames() const { return _decayKinematicsParticleNames; }
 		const std::vector<std::string>& additionalSavedVariableLables() const { return _additionalSavedVariableLabels; }
+		const std::map<std::string,double>& additionalSavedValues() const { return _additionalSavedValues; }
+		double getAdditionalSavedValue(const std::string& name) const { return _additionalSavedValues.at(name); }
+		void setAdditionalSavedValue(const std::string& name, const double value) { _additionalSavedValues[name] = value; }
+		bool hasAdditionalSavedValue(const std::string& name) const { return _additionalSavedValues.find(name) != _additionalSavedValues.end(); }
```
suhlatwork: you could shorten this to `_additionalSavedValues.count(name) > 0`.

storageFormats/eventMetadata.cc
```diff
@@ -316,6 +316,22 @@ eventMetadata* rpwa::eventMetadata::merge(const vector<const eventMetadata*>& in
...
 		mergee->setBinningMap(mergedBinningMap);
 	}
+
+	for (const auto& input : inputData) {
+		for (const auto& nameValue : input->additionalSavedValues()) {
+			if (not mergee->hasAdditionalSavedValue(nameValue.first)) {
+				mergee->setAdditionalSavedValue(nameValue.first, nameValue.second);
+			} else { // the additional value exists in multiple files -> only merge if it is the same
+				if (mergee->getAdditionalSavedValue(nameValue.first) != nameValue.second) {
+					printWarn<< "additional value '" << nameValue.first << "' has different values in the different files." << endl;
```
suhlatwork: please add a space between `printWarn` and `<<`.

suhlatwork: also `printWarn` should be `printErr` as this errors appears fatal to me.

suhlatwork: but apparently we never use `printErr` in this file, so never mind.

storageFormats/eventMetadata.cc
```diff
@@ -316,6 +316,22 @@ eventMetadata* rpwa::eventMetadata::merge(const vector<const eventMetadata*>& in
 	if(mergeDiffMeta) {
 		mergee->setBinningMap(mergedBinningMap);
 	}
+
+	for (const auto& input : inputData) {
```
stefan-wallner: Thanks a lot for your comment.
I fully agree that one should be cautious with using `auto`, but in the context of for loops I thinks it is very useful, because you usually spare lots of characters, which makes the code much more readable.

storageFormats/eventMetadata.h
stefan-wallner: This would be shorter, but I think the semantic for find is more clear, because here we want to find whether the map has that key, not count it. 
Performance wise it does not matter.

pyInterface/bindings/storageFormats/eventMetadata_py.cc
```diff
@@ -86,6 +86,9 @@ void rpwa::py::exportEventMetadata() {
 		.def("productionKinematicsParticleNames", &eventMetadata_productionKinematicsParticleNames)
 		.def("decayKinematicsParticleNames", &eventMetadata_decayKinematicsParticleNames)
 		.def("additionalSavedVariableLables", &eventMetadata_additionalSavedVariableLables)
```
bgrube: Not your fault, but it would be nice to fix typo: "additionalSavedVariableLables" -> "additionalSavedVariableLabels"

pyInterface/bindings/storageFormats/eventMetadata_py.cc
```diff
@@ -86,6 +86,9 @@ void rpwa::py::exportEventMetadata() {
...
+		.def("getAdditionalSavedValue", &rpwa::eventMetadata::getAdditionalSavedValue)
```
bgrube: In ROOTPWA we usually not decorate getters with "get", so it would be nice to rename "getAdditionalSavedValue" -> "additionalSavedValue"

storageFormats/eventMetadata.cc
```diff
@@ -316,6 +316,22 @@ eventMetadata* rpwa::eventMetadata::merge(const vector<const eventMetadata*>& in
 	if(mergeDiffMeta) {
 		mergee->setMultibinBoundaries(mergedmultibinBoundaries);
 	}
+
+	for (const auto& input : inputData) {
```
bgrube: Why make a separate loop here? Why not include it in the loop over the input data above?

suhlatwork:
Updated with respect to 40e7594900237153d7b9780c2152e7b3f3d50fb7: replaced more occurrences of `userString` with `auxString`, style.

suhlatwork:
Merged, thank you!

PR #216, opened 2017-06-03
partialWaveFit: fix compiler warning for non-instantiated variable

suhlatwork:
The 'pwaLikelihood<complexT>::_debug' variable needs to be
instantiated to be accessible from '...::setQuiet'. The variable is
only declared in the header file, so move the function definiation to
the implementation file. (-Wundefined-var-template)

Appears to be a new check in Apple LLVM version 8.1.0.


bgrube:
Merged. Thanks you!

PR #218, opened 2017-06-09
Get fit results

stefan-wallner:
C++ implementation of getFitResultsFromFiles.

### DIFF
--- a/partialWaveFit/fitResult.cc
+++ b/partialWaveFit/fitResult.cc
@@ -44,6 +44,8 @@
 #include "TMath.h"
 #include "TMatrixDSym.h"
 #include "TRandom3.h"
+#include "TFile.h"
+#include "TTree.h"
 
 #include "fitResult.h"
 
@@ -89,26 +91,22 @@ fitResult::fitResult()
 { }
 
 
-fitResult::fitResult(const fitResult& result)
-	: TObject(),
-	  _nmbEvents             (result.nmbEvents()),
-	  _normNmbEvents         (result.normNmbEvents()),
-	  _multibinBoundaries    (result.multibinBoundaries()),
-	  _logLikelihood         (result.logLikelihood()),
-	  _rank                  (result.rank()),
-	  _prodAmps              (result.prodAmps()),
-	  _prodAmpNames          (result.prodAmpNames()),
-	  _waveNames             (result.waveNames()),
-	  _covMatrixValid        (result.covMatrixValid()),
-	  _fitParCovMatrix       (result.fitParCovMatrix()),
-	  _fitParCovMatrixIndices(result.fitParCovIndices()),
-	  _normIntegral          (result.normIntegralMatrix()),
-	  _acceptedNormIntegral  (result.acceptedNormIntegralMatrix()),
-	  _normIntIndexMap       (result.normIntIndexMap()),
-	  _phaseSpaceIntegral    (result.phaseSpaceIntegralVector()),
-	  _converged             (result.converged()),
-	  _hasHessian            (result.hasHessian())
-{ }
+fitResult::fitResult(const fitResult& result, const bool fillCovMatrix, const bool fillIntegralMatrices)
+	: TObject()
+{
+	fill(result, fillCovMatrix, fillIntegralMatrices);
+}
+
+fitResult::fitResult(const fitResult&                          result,
+                     const TMatrixT<double>*                   fitParCovMatrix,         // covariance matrix of fit parameters
+                     const std::vector<std::pair<int, int> >*  fitParCovMatrixIndices,  // indices of fit parameters for real and imaginary part in covariance matrix
+                     const rpwa::complexMatrix*                normIntegral,            // normalization integral matrix
+                     const rpwa::complexMatrix*                acceptedNormIntegral,    // normalization integral matrix with acceptance
+                     const std::vector<double>*                phaseSpaceIntegral)     // normalization integral over full phase space without acceptance
+	: TObject()
+{
+	fill(result, fitParCovMatrix, fitParCovMatrixIndices, normIntegral, acceptedNormIntegral, phaseSpaceIntegral);
+}
 
 
 fitResult::~fitResult()
@@ -191,16 +189,15 @@ fitResult::fill(const unsigned int              nmbEvents,               // numb
                 const int                       rank,                    // rank of fit
                 const vector<complex<double> >& prodAmps,                // production amplitudes
                 const vector<string>&           prodAmpNames,            // names of production amplitudes used in fit
-                const TMatrixT<double>&         fitParCovMatrix,         // covariance matrix of fit parameters
-                const vector<pair<int, int> >&  fitParCovMatrixIndices,  // indices of fit parameters for real and imaginary part in covariance matrix matrix
-                const complexMatrix&            normIntegral,            // normalization integral matrix
-                const complexMatrix&            acceptedNormIntegral,    // normalization integral matrix with acceptance
-                const vector<double>&           phaseSpaceIntegral,      // normalization integral over full phase space without acceptance
+                const TMatrixT<double>*         fitParCovMatrix,         // covariance matrix of fit parameters
+                const vector<pair<int, int> >*  fitParCovMatrixIndices,  // indices of fit parameters for real and imaginary part in covariance matrix matrix
+                const complexMatrix*            normIntegral,            // normalization integral matrix
+                const complexMatrix*            acceptedNormIntegral,    // normalization integral matrix with acceptance
+                const vector<double>*           phaseSpaceIntegral,      // normalization integral over full phase space without acceptance
                 const bool                      converged,
                 const bool                      hasHessian)
 {
 	_converged          = converged;
-	_hasHessian         = hasHessian;
 	_nmbEvents          = nmbEvents;
 	_normNmbEvents      = normNmbEvents;
 	_multibinBoundaries = multibinBoundaries;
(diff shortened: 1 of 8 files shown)


pyInterface/bindings/partialWaveFit/fitResult_py.cc
```diff
@@ -304,6 +333,42 @@ namespace {
...
+		                bp::extract<std::string>(branchName),
+		                bp::extract<bool>(onlyBestInMultibin),
+		                bp::extract<bool>(stripMatricesFromFurtherAttempts),
+		                bp::extract<bool>(onlyConvergedResults));
+
+		bp::object mod = bp::import("pyRootPwa.utils");
+
+		for (auto& binninbMap_fitResultsC : fitResultsInMultibinsC) {
+			bp::list results;
+			for (auto& fitResult : binninbMap_fitResultsC.second) {
```
stefan-wallner: remove the element from `binninbMap_fitResultsC.second` after adding it to `results` to avoid keeping the fit results twice in memory.

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,149 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+
+void fitResult::stripIntegralMatrices() {
+	_normIntegral.resizeTo(0,0);
+	_acceptedNormIntegral.resizeTo(0,0);
+	_phaseSpaceIntegral.clear();
+	_normIntIndexMap.clear();
+}
+
+
+std::map<rpwa::multibinBoundariesType, std::list<rpwa::fitResult> > rpwa::getFitResultsFromFilesInMultibins(
```
suhlatwork: why return a `list` of fit results, not a `vector`? Internally the usage of a `list` reduces the number of copies of fit result objects created, and simplifies a number of things. But for external use it does not allow direct access to specific elements. Converting to a python object requires an extra step. Are the internal advantages really worth the external hassles?

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,149 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+
+				// if the best one is not converged and we want to store all and we want to strip the matrices from all except the best ones (converged and not converged)
+				if( negLogLikeOfBestConvergedResult.find(multibinBoundaries) != negLogLikeOfBestConvergedResult.end()){ // found at least one converged result
+					if (not iterOfBestResult[multibinBoundaries]->converged()){
+						// load best converged result
+						trees[fileEntryOfBestConvergedResult[multibinBoundaries].first]->GetEntry(fileEntryOfBestConvergedResult[multibinBoundaries].second);
+						iterOfBestConvergedResult[multibinBoundaries]->reset();
+						iterOfBestConvergedResult[multibinBoundaries]->fill(fitResultIn);
+					} else {
+						assert(negLogLikeOfBestResult[multibinBoundaries] == negLogLikeOfBestConvergedResult[multibinBoundaries]);
```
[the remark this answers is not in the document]
stefan-wallner: What do you mean by indices?

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,149 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+}
+
+
+std::map<rpwa::multibinBoundariesType, std::list<rpwa::fitResult> > rpwa::getFitResultsFromFilesInMultibins(
+                                                                                                      const std::vector<std::string>& fileNames,
+                                                                                                      const std::string& treeName,
+                                                                                                      const std::string& branchName,
+                                                                                                      const bool onlyBestInMultibin,
+                                                                                                      const bool stripMatricesFromFurtherAttempts,
+                                                                                                      const bool onlyConvergedResults) {
```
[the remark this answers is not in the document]
stefan-wallner: I would avoid changing the order here. This will break my code in lots of points.
But we can make default values for the later arguments as well. e.g.
`onlyBestInMultibin = false`
`stripMatricesFromFurtherAttempts = false`
`onlyConvergedResults = true`

pyInterface/bindings/partialWaveFit/fitResult_py.cc
```diff
@@ -320,6 +323,44 @@ namespace {
...
+
+		std::map<rpwa::multibinBoundariesType, std::list<rpwa::fitResult> > fitResultsInMultibins =
+		        rpwa::getFitResultsFromFilesInMultibins(fileNamesC,
+		                                                bp::extract<std::string>(treeName),
+		                                                bp::extract<std::string>(branchName),
+		                                                bp::extract<bool>(onlyBestInMultibin),
+		                                                bp::extract<bool>(stripMatricesFromFurtherAttempts),
+		                                                bp::extract<bool>(onlyConvergedResults));
+
+		bp::object mod = bp::import("pyRootPwa.utils");
```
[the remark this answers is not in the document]
stefan-wallner: I implemented now a dedicated function for this.
What we have in C++ is  not a `Multibin` object but just the boundaries part of a `Multibin` object, which is a `dict`.
Therefore, I think we should distinguish between these two cases.

pyInterface/module/utils/_fitTreeUtils.py
```diff
@@ -36,43 +36,52 @@ def getFitResultFromFile(fitResultFileName,
...
+                              fitResultTreeName = "pwa",
+                              fitResultBranchName = "fitResult_v2"
+                             ):
 
-	return bestResults
+	bestResultsInMultibin = pyRootPwa.core.getFitResultsFromFilesInMultibins([fitResultFileName], fitResultTreeName, fitResultBranchName, True, False, True)
+	multiBins = set([ b.getSubMultiBin(exclude="mass") for b in bestResultsInMultibin.keys()])
+	if len(multiBins) != 1:
+		printWarn("Fit result file '{0}' contains more than one bin in a variable other than mass: {1}".format(fitResultFileName, list(multiBins)))
+	return {b.getBinCenters()["mass"]: results[0] for b,results in bestResultsInMultibin.iteritems()}
```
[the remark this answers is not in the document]
stefan-wallner: I agree in principle.
However, this functions is not new and the old definition was, that it returns a dict with the bin mass-centers as keys.
Changing this would break quite some code, especially also 'end-user' code.
I think we should avoid changing the 'end-user' interface, if there is not a very strong point to do so.

partialWaveFit/fitResult.cc
stefan-wallner: The main reason was, that the sorting should be more efficient, as the fit results are big objects and also the iterators I use above stay valid after inserting a new element, which is not the case for a vector (see https://stackoverflow.com/questions/6438086/iterator-invalidation-rules).

Internally, I would stick to the `list` type, anyway.
The only use case I see so far is the one via the Python interface.
So I do not think it is rally advantageous to make here a conversion to `vector`, because we do the conversion than in the Python interface.

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,149 @@ fitResult::intensityErr(const string& waveNamePattern) const
 	const TMatrixT<double> intensityCov = jacobian * prodAmpCovJT;              //  1 x  1 matrix
 	return (double)normNmbEvents() * sqrt(intensityCov[0][0]);
 }
+
+
+void fitResult::stripCovMatrix() {
```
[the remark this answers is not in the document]
stefan-wallner: I have one use case for this two functions.
In a later pull request :), I add to the `pwaFit` script the possibility to store only the covariance and integral matrices of the best fit result of all attempts.
This requires the possibility to strip the matrices afterwards.

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,151 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+}
+
+
+std::map<rpwa::multibinBoundariesType, std::list<rpwa::fitResult> >
+rpwa::getFitResultsFromFilesInMultibins(
+                                        const std::vector<std::string>& fileNames,
+                                        const std::string& treeName,
+                                        const std::string& branchName,
+                                        const bool onlyBestInMultibin,
+                                        const bool stripMatricesFromFurtherAttempts,
```
bgrube: It would be nice to make the "Further" more explicit. Does i t mean "NonBest"?

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,151 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+                                        const bool onlyConvergedResults) {
```
bgrube: The naming of the three bool flags is somewhat inconsistent: Attempt vs. Result vs. onlyBest<what?>InMultibin

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,151 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+	std::map<rpwa::multibinBoundariesType, std::list<rpwa::fitResult> > fitResultsInMultibins;
+	std::unique_ptr<fitResult> fitResultIn(new fitResult);
```
bgrube: The "In" is ambiguous. I guess it stands for input. If yes, then "inputFitResult" would be better.

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,151 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+		tree->SetBranchAddress(branchName.c_str(), fitResultIn.get());
+		trees.push_back(tree);
+		files.push_back(file);
+
+	}
+
+	// find the best result in each multibin and store all others if necessary
+	std::set<rpwa::multibinBoundariesType> allMultibinBoundaries;
+	std::map<rpwa::multibinBoundariesType, double> negLogLikeOfBestResult;
+	std::map<rpwa::multibinBoundariesType, std::pair<size_t, int> > fileEntryOfBestResult;
```
bgrube: maybe "fileTreeEntry" would be more helpful?

partialWaveFit/fitResult.h
```diff
@@ -582,7 +586,22 @@ namespace rpwa {
 		}
 		return prodAmpName.substr(prodAmpName.find('_')+1);
 	}
-
+	/// \brief returns list of fit results loaded from the given files
```
bgrube: returns for each kinematic bin...

partialWaveFit/fitResult.cc
stefan-wallner: Yes, the "In" in fitResultsInMultibins means that the fit results are grouped into multibins.

partialWaveFit/fitResult.cc
```diff
@@ -998,3 +1000,151 @@ fitResult::intensityErr(const string& waveNamePattern) const
...
+	// finally add the result with the best likelihood, if not yet stored
+	if (onlyBestInMultibin) {
+		for (const auto& multibinBoundaries : allMultibinBoundaries) {
+			// load best result
+			trees[fileEntryOfBestResult[multibinBoundaries].first]->GetEntry(fileEntryOfBestResult[multibinBoundaries].second);
+			std::list<fitResult>& fitResults = fitResultsInMultibins[multibinBoundaries];
+			assert(fitResults.size() == 0);
+			fitResults.push_back(*fitResultIn);
+		}
+	} else if (stripMatricesFromFurtherAttempts) {
```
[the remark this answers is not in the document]
stefan-wallner: In principle, I agree.
From my point of view, there are two possibilities to realize this and both have serious drawbacks.
Either you first load all fit results with their matrices into memory, than determine the best one and finally strip the matrices from all other results. However, this would mean that you need enough memory to keep all fit results with their matrices of all attempts from all multibins (if you want collect your results e.g.) in memory, which might be quite a lot.
Or you have to go twice over all fit result files. Fist to determine the best results and second to load the results, while keeping the matrices only for the best results. Going twice over all files is also not really a good solution.

bgrube:
Merged. Thanks a lot!

PR #220, opened 2017-06-13
Fixes and improvements

suhlatwork:
Various fixes and improvements mainly in the Python interface.

Most important is probably the fix to make building/linking on MacOS work again.


bgrube:
Merged. Thank you!

PR #222, opened 2017-06-19
Python bindings for resonance fit

suhlatwork:
This pull request adds the Python bindings for the resonance model fit.

As the smart pointers from C++11 require Boost 1.63.0 to be allowed in the Python interface, Boost again is 'manually' installed in Travis, to save some time, this pull request attempts to use the caches provided by Travis. Let's see if this works.


bgrube:
ab731a35078583d766635b51189f2b5f9fc2c78f included in #241.

PR #223, opened 2017-06-22
Naming of multibin boundaries

stefan-wallner:
Naming the object storing the variables and boundaries of a multi-bin 'multibinBoundaries' in the whole code.
Especially, 'binningMap' was renamed to 'multibinBoundaries'.
Additionally, naming the corresponding type 'multibinBoundariesType' and replacing the explicit type declaration 'std::map<...' by 'multibinBoundariesType'.


suhlatwork:
Merged as commit 0c6259e82f6ad38d63135fb0fa8e109472a8c64b with mostly cosmetic changes:
* moved the `typedef`s for the boundary types into its own header file
* replaced even more of the container massacre with the `typedef`s

Thank you!

PR #224, opened 2017-06-27
storageFormats: additionalTreeVariables class

stefan-wallner:
Implement additionalVariables class that handles additional
variables of an event tree.
Additionally, impelement a member function that uses these variables
to determine wether the current event is in a given multibin.

### DIFF
--- a/storageFormats/eventMetadata.cc
+++ b/storageFormats/eventMetadata.cc
@@ -365,3 +365,48 @@ std::string rpwa::eventMetadata::getStringForEventsType(const eventsTypeEnum& ty
 	}
 	return "UNKNOWN";
 }
+
+
+bool
+additionalTreeVariables::setBranchAddresses(const eventMetadata& metaData)
+{
+	TTree* tree = metaData.eventTree();
+
+	if (tree != nullptr) {
+		for (const auto& name: metaData.additionalSavedVariableLables()) {
+			int err = tree->SetBranchAddress(name.c_str(), &(_additionalTreeVariables[name]));
+			if (err < 0){
+				printErr << "could not set branch address for branch '" << name << "' (error code " << err << ")." << endl;
+				_additionalTreeVariables.clear();
+				return false;
+			}
+		}
+	} else {
+		printErr << "no tree in eventMetadata object." << std::endl;
+		return false;
+	}
+
+	return true;
+}
+
+
+bool
+additionalTreeVariables::inBoundaries(const multibinBoundariesType& boundaries) const
+{
+	for (const auto& elem: boundaries) {
+		const std::string& name = elem.first;
+		const boundaryType& range = elem.second;
+		const std::map<std::string, double>::const_iterator it = _additionalTreeVariables.find(name);
+		if (it != _additionalTreeVariables.end()) {
+			const double variable = it->second;
+			if (variable < range.first or variable >= range.second) {
+				return false;
+			}
+		} else {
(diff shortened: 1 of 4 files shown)


storageFormats/eventMetadata.cc
```diff
@@ -365,3 +365,45 @@ std::string rpwa::eventMetadata::getStringForEventsType(const eventsTypeEnum& ty
...
+
+
+bool
+additionalVariables::setBranchAddresses(const eventMetadata& metaData)
+{
+	TTree* tree = metaData.eventTree();
+	if (tree != nullptr){
+		for(const auto& label: metaData.additionalSavedVariableLables()){
+			_labels.push_back(label);
+			int err = tree->SetBranchAddress(label.c_str(), &_additionalSavedVariables[label]);
```
bgrube: It may help the reader to add parantheses
`&(_additionalSavedVariables[label])`

storageFormats/eventMetadata.h
```diff
@@ -12,6 +12,9 @@ class TTree;
 
 namespace rpwa {
 
+	class additionalVariables;
```
bgrube: The naming needs to be adapted to merge request #213

storageFormats/eventMetadata.h
```diff
@@ -112,6 +115,31 @@ namespace rpwa {
...
+		/***
+		 * \brief checks wether the current additional variables are within the boundaries
+		 *
+		 * x_i in [lower_i, upper_i) for all variables i in the given boundaries
+		 */
+		bool inBoundaries(const rpwa::multibinBoundariesType& multibinBoundaries)const;
+
+	private:
+		std::map<std::string, double > _additionalSavedVariables;
+		std::vector<std::string> _labels;
```
bgrube: Is it necessary to duplicate the information already stored in `_additionalSavedVariables`?

storageFormats/eventMetadata.h
```diff
@@ -112,6 +115,31 @@ namespace rpwa {
 		return metadata.print(out);
 	}
 
+
+	/// \brief Handles additional saved variables of event data
```
bgrube: maybe one could elaborate a bit more:
class represents kinematic variables used for binning and cutting
provides interface to connect kinematic variables to tree branches and to check whether event is within a kinematic cell

storageFormats/eventMetadata.h
```diff
@@ -112,6 +115,31 @@ namespace rpwa {
...
+	class additionalVariables {
+	public:
+		additionalVariables() {}
+		additionalVariables(const additionalVariables&) = delete;
+		additionalVariables& operator = (const additionalVariables&) = delete;
+		double operator [](const std::string& label) const {return _additionalSavedVariables.at(label);}
+
+		bool setBranchAddresses(const eventMetadata& metaData);
```
bgrube: Add description:
reads list of tree variables from provided metaData and connects them to tree branches

suhlatwork:
With respect to 91503d6f573c88b16e4334b2e1c03f3965f55d08 removed some now unused variables, and improved the coding style.

Do we need the `additionalTreeVariables::labels` function? This seems to be not used, it is not exported to Python, and I do not think it is useful to get it from this class. After all an object of this class can only be brought to a useful state with an `eventMetadata` object, so one could always query the `eventMetadata` object for this information.

stefan-wallner:
I agree that `additionalTreeVariables::labels` is not needed. I removed it.

suhlatwork:
Merged. Thank you!

PR #225, opened 2017-06-29
pyInterface: Implement Python converter for multibinBoundariesType

stefan-wallner:
Implement a converter functions between Python and C++ for the
multibinBoundariesType.
Replace the multiple implementations of this conversion by the common
converter functions.


suhlatwork:
Please check [branch](https://github.com/suhlatwork/ROOTPWA/commit/6e4a6b065b5677449890f4240eaf3aa4ca1c80ec). By adding the converters to the Boost.Python interface we would no longer need to take care to call the proper functions on the C++ side, but could simply write the type we want.

stefan-wallner:
I agree that this is a very elegant way.
However, with your proposal we introduce a general converter from `map<string, pair<double,double>>` to a python dictionary of (string, tuple), not just for the special case of a binning map.
I'm not sure whether this is a good idea.
Alternatively, we could implement your proposal and define 'multibinBoundariesType' as class, not just a typedef.

suhlatwork:
Merged, thank you!

PR #226, opened 2017-06-29
pyInterface: Additional functionality of the multiBin class

stefan-wallner:
Implement hash function for the multiBin class, such that
it can be used as key in a dict.
Additionally, the printing of the multiBin class is improved.
Implement fetching of a sub-bin.
Implement 'A in B' operator if bin A is contained in bin B.
Remove restriction that a multiBin must have at least one variable
(this restriction, if needed, should be implemented in the config file)
Remove constrain that a the boundaries of a multi-bin cannot be equal,
because this is needed if we have only the bin-center information
available (e.g. old version of fitResult class).


pyInterface/module/utils/_binning.py
```diff
@@ -61,15 +63,94 @@ def __ge__(self, other):
...
-			retval += "\"" + key + "\": (" + str(self.boundaries[key][0]) + ", " + str(self.boundaries[key][1]) + "), "
-		retval = retval[:-2] + " }"
+			retVariables.append( "\"" + key + "\": (" + str(self.boundaries[key][0]) + ", " + str(self.boundaries[key][1]) + ")")
+		retval = "{ "
+		retval += ", ".join(retVariables)
+		retval += " }"
 		return retval
 
+	def __hash__(self):
+		keys = tuple(self.boundaries.keys())
```
suhlatwork: Is it ensured that `keys` returns a sorted list? My understanding of the Python reference is that it is not, so we should probably do it ourselves, i.e., use `tuple(self.variables())`

pyInterface/module/utils/_binning.py
```diff
@@ -1,3 +1,5 @@
+import copy
+from pyRootPwa.utils._printingUtils import printErr
 from _printingUtils import printDebug
```
suhlatwork: Combine those two line to one.

pyInterface/module/utils/_binning.py
```diff
@@ -61,15 +63,94 @@ def __ge__(self, other):
...
+		If this multiBin is binned in a variable in which the other multiBin is not binned,
+		the other multiBin is not contained in this multiBin.
+		If the other multiBin is binned in a variable in which this multiBin is not binned,
+		the other multiBin is contained in this multiBin.
+		'''
+		if isinstance(other, dict):
+			other = multiBin(other)
+		if not isinstance(other, multiBin):
+			printErr("Cannot check whether object of type {0} is contained in a multiBin".format(type(other)))
+			exit(1)
```
suhlatwork: Throwing an exception would be a more Pythonic way.

pyInterface/module/utils/_binning.py
```diff
@@ -61,15 +63,94 @@ def __ge__(self, other):
...
+		'''
+		return sorted(self.boundaries.keys())
+
+	def uniqueStr(self):
+		'''
+		@return: string, unique for the multibin (without whitespaces)
+		'''
+		out = []
+		for variable in self.variables():
+			variableStr = "{0}_{1!r}_{2!r}".format(variable, self.boundaries[variable][0], self.boundaries[variable][1])
```
suhlatwork: Personally, I would not use any character here that is allowed in a valid C++ variable name.

pyInterface/module/utils/_binning.py
```diff
@@ -61,15 +63,94 @@ def __ge__(self, other):
...
+				else:
+					printErr("Cannot get multiBin form string '{0}'".format(strIn))
+		return multiBin(boundaries)
 
 	def overlap(self, other, strict=True):
+		'''
+		@return: True if this and the other multiBin overlap in all variables, False otherwise.
+		'''
+		keys = sorted(self.boundaries.keys())
+		if keys != sorted(other.boundaries.keys()):
```
suhlatwork: you could use the `variables` function

pyInterface/module/utils/_binning.py
```diff
@@ -78,17 +159,12 @@ def comparator(left, right, direction):
...
+		if comparator(self.boundaries[key][0], other.boundaries[key][1], ">"):
 			return False
-		for key in keys:
-			if comparator(self.boundaries[key][1], other.boundaries[key][0], "<"):
-				return False
-			if comparator(self.boundaries[key][0], other.boundaries[key][1], ">"):
-				return False
 		return True
 
-
```
suhlatwork: this empty should not be removed

pyInterface/module/utils/_binning.py
```diff
@@ -102,6 +178,26 @@ def inBin(self, binningInfo):
...
+		if isinstance(exception, str):
+			exception = [exception]
+		boundaries = {}
+		for k,values in self.boundaries.iteritems():
+			if k not in exception:
+				boundaries[k] = values
+		return multiBin(boundaries)
+
+	def getBinCenters(self):
+		return {k: 0.5*(v[0]+v[1]) for k,v in self.boundaries}
```
suhlatwork: should be `self.boundaries.iteritems()`

PR #229, opened 2017-10-09
decayAmplitude: Parameterizations for [Kpi]_S wave

stefan-wallner:
Implement generic GLASS parameterization from LHCb/BaBar
and K-Matrix parameterization from Palano, Pennington.
Implement possibility to handel parameters for the mass-dependent isobar
parameterization from the key file to the parameterization.

### DIFF
--- a/decayAmplitude/massDependence.cc
+++ b/decayAmplitude/massDependence.cc
@@ -34,11 +34,12 @@
 #include "massDependence.h"
 
 #include <type_traits>
+#include <regex>
 
 #include <boost/numeric/ublas/io.hpp>
 
-#include "isobarDecayVertex.h"
 #include "libConfigUtils.hpp"
+#include "isobarDecayVertex.h"
 #include "particleDataTable.h"
 #include "phaseSpaceIntegral.h"
 #include "physUtils.hpp"
@@ -67,10 +68,6 @@ namespace {
 	template<typename T>
 	typename std::enable_if<not std::is_default_constructible<T>::value, rpwa::massDependencePtr>::type createMassDependence(const libconfig::Setting* settings)
 	{
-		if (settings == nullptr) {
-			printErr << "mass dependence '" << T::Name() << "' requires additional settings, but none are specified. Aborting..." << std::endl;
-			throw;
-		}
 		return T::Create(settings);
 	}
 
@@ -93,6 +90,21 @@ namespace {
 	}
 
 
+	/**
+	 * Check if the configuration with the given name exists in the settings.
+	 * If it does not exist or cannot be converted to type `T`, print a warning and throw an exception.
+	 * @param massDependenceName Name of the mass dependence implementation
+	 * @param setting Setting to look up
+	 * @param name Name of the variable in the setting
+	 * @param valueOut Output variable, the value will be written to
+	 */
+	template<typename T>
+	static void lookupParameter(const string& massDependenceName, const libconfig::Setting* setting, const std::string& parameter, T& valueOut) {
+		if (not setting->lookupValue(parameter, valueOut)) {
+			printErr<< "cannot find '" << parameter << "' value in configuration for mass dependence '" << massDependenceName << "'. Aborting..." << std::endl;
+			throw;
+		}
+	}
 }
 
 
@@ -126,13 +138,14 @@ rpwa::createMassDependence(const std::string& massDepType, const libconfig::Sett
 	                              rpwa::piPiSWaveAuMorganPenningtonM,
 	                              rpwa::piPiSWaveAuMorganPenningtonVes,
 	                              rpwa::piPiSWaveAuMorganPenningtonKachaev,
-	                              rpwa::rhoPrimeMassDep>(massDepType, settings);
+	                              rpwa::rhoPrimeMassDep,
+	                              rpwa::KPiSGLASS,
+	                              rpwa::KPiSPalanoPennington>(massDepType, settings);
 }
 
 
 ////////////////////////////////////////////////////////////////////////////////
 bool massDependence::_debug = false;
-
 
 std::string
 massDependence::parentLabelForWaveName(const isobarDecayVertex& v) const
@@ -176,6 +189,11 @@ flatMassDependence::parentLabelForWaveName(const isobarDecayVertex& v) const
 boost::shared_ptr<binnedMassDependence>
 binnedMassDependence::Create(const libconfig::Setting* settings)
 {
+	if (settings == nullptr) {
+		printErr << "no settings given for binned mass dependence." << std::endl;
+		throw;
+	}
+
 	const libconfig::Setting* bounds = rpwa::findLibConfigList(*settings, "bounds" , false);
 	if (not bounds) {
 		printErr << "no bounds given for binned mass dependence." << std::endl;
@@ -416,7 +434,7 @@ f0980FlatteBesII::amp(const isobarDecayVertex& v)
 
 	denom -= imag * g1 * (rhoPi + g2g1*rhoK);
 
-	const std::complex<double> amp(1. / denom);
+	const std::complex<double> amp = 1./denom;
 
 	if (_debug)
 		printDebug << name() << "(m = " << maxPrecision(M) << " GeV/c^2, m_0 = " << maxPrecision(M0)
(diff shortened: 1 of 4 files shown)


suhlatwork:
Please consider rebasing to and using suhlatwork/ROOTPWA@f1b846abfb0a3b804e164f72a04846327611cdae.

In general you should keep in mind that the `waveSetGenerator` cannot handle mass dependencies with additional parameters (yet(?)).

* I would rename the classes to `kPiSGlass` and `kPiSPalanoPennington` (or `KPiSGlass` and `KPiSPalanoPennington` if the K has to be in capitals) to have all mass dependence classes in camel case.
* there should be references to the respective publications
* what is the reason for the additional `ampM` methods? If it is just for plotting, we have `getMassShapes` in `highLevelInterface` to get also the individual two-body/isobar amplitudes, @fkrinner should be able to provide an example of how to use it.

decayAmplitude/massDependence.cc
```diff
@@ -31,13 +31,19 @@
...
 
 
+#include "massDependence.h"
+
+#include <type_traits>
+
 #include <boost/numeric/ublas/io.hpp>
 
+#include "libConfigUtils.hpp"
 #include "physUtils.hpp"
```
bgrube: remove here; is already included in #L46

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
 }
 
 
+////////////////////////////////////////////////////////////////////////////////
+KPiSGLASS::KPiSGLASS(const double a, const double r, const double M0, const double G0, const double phiF, const double phiR, const double phiRsin,
+                     const double F, const double R, const double MMax, const std::string& tag)
```
bgrube: Give equation number in paper.
It would be nice to briefly explain meaning of parameters, if possible

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+	  _F(F),
+	  _R(R),
+	  _MMax(MMax),
+	  _tag(tag),
+	  _piChargedMass(0.0),
+	  _kaonChargedMass(0.0)
+{
+	particleDataTable& pdt = particleDataTable::instance();
+	const string partList[] = { "pi+", "K+" };
+	for (unsigned int i = 0; i < sizeof(partList) / sizeof(partList[0]); ++i)
```
bgrube: Maybe size_t would be better.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+
+		if(not massDepKey->lookupValue("tag", tag)){
+			printErr << "cannot find 'tag' value in configuration for mass dependence '" << Name() << "'. Aborting..." << endl;
+			throw;
+		}
+
+		return KPiSGLASS::Create(a, r, M0, G0, phiF, phiR, phiRsin, F, R, MMax, tag);
+	} else {
+		printWarn << "no configuration given for mass dependence '" << Name() << "'. Using dummy config" << endl;
+		return KPiSGLASS::Create(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, "");
```
bgrube: Are there meaningful default parameter values?

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+	}
+}
+
+
+std::string
+KPiSGLASS::parentLabelForWaveName(const isobarDecayVertex& v) const{
+	std::ostringstream label;
+	label << name() << "<" << _tag << ">[" << v.parent()->name() << "]";
```
bgrube: Should we include the parameter values?

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+	return label.str();
+}
+
+
+complex<double>
+KPiSGLASS::amp(const isobarDecayVertex& v)
```
bgrube: Same comments as for KPiSGLASS::KPiSGLASS() in #L621 ff.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		ampl *= M/q; // divide by Kpi phase-space
+	}
+
+	return ampl;
+
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+KPiSPalanoPennington::KPiSPalanoPennington(const double MMax)
```
bgrube: Same comments as for KPiSGLASS::KPiSGLASS() in #L621 ff.

decayAmplitude/massDependence.h
```diff
@@ -94,7 +102,7 @@ namespace rpwa {
 	/// create a mass dependence object as specified by 'massDepType'
 	// only marks for mass dependencies not requiring additional arguments,
```
bgrube: "marks" -> "works"?
the sentence seems to be not true anymore

decayAmplitude/massDependence.h
```diff
@@ -415,6 +426,75 @@ namespace rpwa {
...
+		~KPiSGLASS()                        { }
+
+		using massDependenceImpl<KPiSGLASS>::Create;
+		static boost::shared_ptr<KPiSGLASS> Create(const libconfig::Setting* settings);
+
+		virtual std::string parentLabelForWaveName(const isobarDecayVertex& v) const;  ///< returns label for parent of decay used in wave name
+
+		virtual std::complex<double> amp(const isobarDecayVertex& v);
+
+		static constexpr const char* cName = "KPiSGLASS";
```
bgrube: general comment: shouldn't it be _cName throughout for consistency?

pyInterface/bindings/decayAmplitude/massDependence_py.cc
```diff
@@ -452,5 +541,7 @@ void rpwa::py::exportMassDependence() {
 	bp::register_ptr_to_python<rpwa::piPiSWaveAuMorganPenningtonVesPtr>();
 	bp::register_ptr_to_python<rpwa::piPiSWaveAuMorganPenningtonKachaevPtr>();
 	bp::register_ptr_to_python<rpwa::rhoPrimeMassDepPtr>();
+	bp::register_ptr_to_python<rpwa::KpiSGLASSPtr>();
+	bp::register_ptr_to_python<rpwa::KpiSPalanoPenningtonPtr>();
```
bgrube: python interface for rpwa::binnedMassDependence is missing

decayAmplitude/massDependence.cc
stefan-wallner: `pdt.entry` does only print a warning and returns `NULL`, so you would need to make a check of that `NULL` in both cases below.

decayAmplitude/massDependence.cc
stefan-wallner: The only reasonable default parameters would the the ones of the simplifies version called LASS, which resulted from a fit to LASS data.
However, there are also more results flying around:
- The original one form the LASS paper
- A refit by Dunwoodie (just some webpage), whose  parameters were used in a recent LHCb analysis
So I would leave it to the user to choose some parameters.

decayAmplitude/massDependence.cc
stefan-wallner: I do not think so, because this would lead to extremely long file names.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +617,231 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+	_kaonChargedMass = pdt.entry("K+")->mass();
+	_etaMass = pdt.entry("eta0")->mass();
+}
+
+
+boost::shared_ptr<KPiSPalanoPennington>
+KPiSPalanoPennington::Create(const libconfig::Setting* massDepKey){
+
+	// if not given, take up to 2.4 GeV
+	double MMax = 2.4;
```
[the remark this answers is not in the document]
stefan-wallner: I have to check whether this breaks the way we decide which `createMassDependence` functions is build

decayAmplitude/massDependence.h
stefan-wallner: I do not think so. 
It is a static **public** member.

decayAmplitude/massDependence.cc
stefan-wallner: Done in the header file

decayAmplitude/massDependence.cc
stefan-wallner: Done in the header file and in the cc file (for the parameters)

pyInterface/bindings/decayAmplitude/massDependence_py.cc
stefan-wallner: Done, two lines above.

decayAmplitude/massDependence.h
```diff
@@ -415,6 +420,77 @@ namespace rpwa {
 	typedef boost::shared_ptr<rhoPrimeMassDep> rhoPrimeMassDepPtr;
 
 
+	//////////////////////////////////////////////////////////////////////////////
+	/// Brief GLASS parameterization for [Kpi]_S wave
+	/// Taken from The European Physical Journal C 2014 vol: 74 (11) pp: 3026
+	/// Equations 13.2.13 to 13.2.15 and 13.2.4
+	class KPiSGLASS: public massDependenceImpl<KPiSGLASS> {
```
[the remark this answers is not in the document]
stefan-wallner: There is a LASS parameterization, and this is the Generalized LASS.
Therefor, I would stick with KPiSGLASS. What do you mean?

pyInterface/bindings/decayAmplitude/massDependence_py.cc
```diff
@@ -386,6 +496,13 @@ void rpwa::py::exportMassDependence() {
 		.def("name", &flatMassDependenceWrapper::name, &flatMassDependenceWrapper::default_name)
 		.def("name", &rpwa::flatMassDependence::name);
 
+	bp::class_<binnedMassDependenceWrapper, bp::bases<rpwa::massDependence> >("binnedMassDependence", bp::init<const double,const double>())
+		.def(bp::self_ns::str(bp::self))
+		.def("amp", &binnedMassDependenceWrapper::amp, &binnedMassDependenceWrapper::default_amp)
+		.def("amp", &rpwa::binnedMassDependence::amp)
+		.def("name", &binnedMassDependenceWrapper::name, &binnedMassDependenceWrapper::default_name)
+		.def("name", &rpwa::massDependence::name);
```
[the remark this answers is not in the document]
stefan-wallner: It should not matter, but I agree that I should be changed for consistency.

stefan-wallner:
What do you mean by

> Please remove the remainders of a merge/rebase conflict in your second commit also from the history

decayAmplitude/massDependence.h
```diff
@@ -415,6 +420,77 @@ namespace rpwa {
...
+
+
+	typedef boost::shared_ptr<KPiSGLASS> KPiSGLASSPtr;
+
+
+	//////////////////////////////////////////////////////////////////////////////
+	/// Brief K-matrix parameterization for [Kpi]_S wave from Palano, Pennington
+	/// Taken from arXiv:1701.04881
+	/// Equations 2 to 6
+	/// Fixed some bugs in the formulas of the original paper
```
suhlatwork: would be nice to precisely point out the corrections, in particular that the second version is not available on the arXiv

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		}
+		std::string forbiddenChars("[\\>\\<\\]\\[\\=]");
+		if (std::regex_search(tag, std::regex(forbiddenChars))){
+			printErr<< "The following characters are forbidden in tags: '" << forbiddenChars << "'" << endl;
+			throw;
+		}
+
+		return KPiSGLASS::Create(a, r, M0, G0, phiF, phiR, phiRsin, F, R, MMax, tag);
+	} else {
+		printErr<< "no configuration given for mass dependence '" << Name() << "'. Using dummy config" << endl;
```
suhlatwork: this should now end with `Aborting...`

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+			throw;
+		}
+
+		if (not massDepKey->lookupValue("tag", tag)) {
+			printErr<< "cannot find 'tag' value in configuration for mass dependence '" << Name() << "'. Aborting..." << endl;
+			throw;
+		}
+		std::string forbiddenChars("[\\>\\<\\]\\[\\=]");
+		if (std::regex_search(tag, std::regex(forbiddenChars))){
+			printErr<< "The following characters are forbidden in tags: '" << forbiddenChars << "'" << endl;
```
suhlatwork: this should end with `Aborting...`

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		const double q0 = breakupMomentum(_M0, _kaonChargedMass, _piChargedMass);
+
+		const double G = _G0 * _M0 / M * q / q0;
+		const double deltaR = atan(_M0 * G / (_M0 * _M0 - M * M));
+
+		const double deltaF = atan(2.0 * _a * q / (2.0 + _a * _r * q * q));
+
+		ampl = _F * sin(deltaF + _phiF) * exp(complex<double>(0.0, deltaF + _phiF))
+		        + _R * sin(deltaR + _phiRsin) * exp(complex<double>(0.0, deltaR + _phiR)) * exp(complex<double>(0.0, 2 * (deltaF + _phiF)));
+		ampl *= M / q; // divide by Kpi phase-space
```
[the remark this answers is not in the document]
stefan-wallner: Depending on the paper, sometimes they skip it, sometimes they keep this factor, as it is just a factor which is normalized out.
In the reference I give, the kept the factor, so I added it.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		const complex<double> rhoKeta = 2 * qKeta / M;
+
+		const double x = (2 * s - (sTop + sBot)) / (sTop - sBot);
+
+		const complex<double> K_11 = (s - sA) / sKpi
+		        * (ga[0] * ga[0] / (sa - s) + gb[0] * gb[0] / (sb - s) + C[0][0][0] + C[0][0][1] * x + C[0][0][2] * x * x + C[0][0][3] * x * x * x);
+		const complex<double> K_12 = (s - sA) / sKpi
+		        * (ga[0] * ga[1] / (sa - s) + gb[0] * gb[1] / (sb - s) + C[0][1][0] + C[0][1][1] * x + C[0][1][2] * x * x + C[0][1][3] * x * x * x);
+		const complex<double> K_22 = (s - sA) / sKpi
+		        * (ga[1] * ga[1] / (sa - s) + gb[1] * gb[1] / (sb - s) + C[1][1][0] + C[1][1][1] * x + C[1][1][2] * x * x + C[1][1][3] * x * x * x);
```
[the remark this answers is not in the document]
stefan-wallner: I think it should be correct. You end up with the correct Breit Wigner if you use this sign.
However, this was one of the things I changed when taking the formula form the paper.

stefan-wallner:
I will to this at the very end, once all your requests are implemented.

> please squash the implement XXX comments with the corresponding original commit

decayAmplitude/massDependence.cc
```diff
@@ -369,13 +420,13 @@ f0980FlatteBesII::amp(const isobarDecayVertex& v)
 
 	denom -= imag * g1 * (rhoPi + g2g1*rhoK);
 
-	const std::complex<double> amp(1. / denom);
+	const std::complex<double> ampl = 1./denom;
```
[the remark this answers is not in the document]
bgrube: Yes, I think this would be nice.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+	_kaonChargedMass = pdt.entry("K+")->mass();
+}
+
+boost::shared_ptr<KPiSGLASS>
+KPiSGLASS::Create(const libconfig::Setting* massDepKey)
+{
+	if (massDepKey != nullptr) {
+		double a, r, M0, G0, phiF, phiR, phiRsin, F, R, MMax;
+		string tag;
+		if (not massDepKey->lookupValue("a", a)) {
```
bgrube: Not sure whether its worth the effort. But have you thought about generating the repetitive 'massDepKey->lookupValue' code using macros?

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		const double deltaR = atan(_M0 * G / (_M0 * _M0 - M * M));
+
+		const double deltaF = atan(2.0 * _a * q / (2.0 + _a * _r * q * q));
+
+		ampl = _F * sin(deltaF + _phiF) * exp(complex<double>(0.0, deltaF + _phiF))
+		        + _R * sin(deltaR + _phiRsin) * exp(complex<double>(0.0, deltaR + _phiR)) * exp(complex<double>(0.0, 2 * (deltaF + _phiF)));
+		ampl *= 0.5*M / q; // divide by Kpi phase-space
+	}
+	if (_debug)
+		printDebug << name() << " M = " << M
```
bgrube: Please use the same log message format as the other amplitudes. See e.g. L585.

decayAmplitude/massDependence.cc
```diff
@@ -564,6 +612,244 @@ piPiSWaveAuMorganPenningtonKachaev::piPiSWaveAuMorganPenningtonKachaev()
...
+		const complex<double> detK = K_11 * K_22 - K_12 * K_12;
+		const complex<double> delta = 1.0 - j * rhoKpi * K_11 - j * rhoKeta * K_22 - rhoKpi * rhoKeta * detK;
+		const complex<double> T_11 = (K_11 - j * rhoKeta * detK) / delta;
+
+		ampl = T_11;
+
+	}
+
+	if (_debug)
+		printDebug << name() << " M = " << M
```
bgrube: Please use the same log message format as the other amplitudes. See e.g. L585.

decayAmplitude/massDependence.cc
suhlatwork: No macros please, but reducing the duplication with a function (of whatever sort) would indeed be a good idea.

bgrube:
Merged. Thank you!

PR #230, opened 2017-10-10
Travis: Fix python problem in travis

stefan-wallner:


suhlatwork:
Thank you! I added a call to `pyenv` to (hopefully) force the usage of system's Python and removed the rest of the previous workaround. Let's see how long it works this time.

PR #234, opened 2018-02-06
cmake: Require new pylint version

stefan-wallner:
Older astroid versions have a problem with numpy, which leads to
problems with some of the pylint chekcs.
Disable the pylint feature if the required astroid version cannot be
found.


suhlatwork:
`pylint` version 1.5.5, which is the minimum version we require in ROOTPWA, requires `astroid` version 1.4.5 as a dependency. Checking for an even older version seems to be odd

pyInterface/module/_fileManager.py
```diff
@@ -17,7 +17,7 @@ def saveFileManager(fileManagerObject, path):
 		return True
 	else:
```
suhlatwork: this `else` could be removed entirely, then it would be more clear that the error message and the `return False` belong together

pyInterface/module/utils/_fitTreeUtils.py
```diff
@@ -1,11 +1,12 @@
...
-
 import pyRootPwa.core
 
+from _printingUtils import printErr, printWarn
 import _root
 ROOT = _root.ROOT
 
 
+
+
```
suhlatwork: why some more empty lines?

pyInterface/CMakeLists.txt
```diff
@@ -162,12 +162,12 @@ if(PYLINT_FOUND)
...
 	add_test(
 		NAME pyLintPyRootPwa
-		COMMAND ${PYLINT_EXECUTABLE} --reports=n --rcfile=${CMAKE_CURRENT_SOURCE_DIR}/pylintrc pyRootPwa
+		COMMAND ${PYLINT_EXECUTABLE} --reports=n --rcfile=${CMAKE_CURRENT_SOURCE_DIR}/pylintrc module
 	)
 	set_property(
 		TEST pyLintPyRootPwa
 		APPEND
-		PROPERTY ENVIRONMENT "PYTHONPATH=${PYTHON_LIBRARY_OUTPUT_PATH}:$ENV{PYTHONPATH}"
+		PROPERTY ENVIRONMENT "PYTHONPATH=${CMAKE_CURRENT_SOURCE_DIR}:${PYTHON_LIBRARY_OUTPUT_PATH}:$ENV{PYTHONPATH}"
```
suhlatwork: It seems we do not need the `${PYTHON_LIBRARY_OUTPUT_PATH}` here at all.

stefan-wallner: Is is required, e.g. if ROOTPWA is not in the PYTHONPATH or another ROOTPWA version is in the PYTHONPATH.

suhlatwork:
Merged in ccfecfe5d1888dd1b33017bea46a57d1d25daf41. Thank you!

Issue #50, opened 2014-08-28
[Feature] Improve handling of flatRange amplitudes [sf#19]


legordian:
**Reported by sebastianuhl on 2013-08-08 16:48 UTC**
A new amplitude 'flatRange' to be able to study, e.g, the pipi S-wave in a 3pi analysis has just been added. However, at the moment the usage is quite cumbersome. To cover the complete two pi mass range, the number of waves to be added is (mass range)/(width of the flatRange) (e.g. for the pipi S-wave with 40 MeV/c2 mass bins this corresponds to 50). For each of those waves
- a new entry has to be made to the particle table (specifying the 2pi mass and a mass range)
- a new keyfile for calcAmplitude has to be created (with the just created particle name)
- calcAmplitude needs to be called, so one amplitude file per wave containing mostly 0
- a new entry has to be added to the wavelist for pwafit with the correct threshold

The long-term goal should probably be
- automatically expand a single keyfile
- to have a single amplitude file containing all the individual (sub-)waves (this will require the amplitudes to be stored in ROOT format)

It is not obvious at which place/program to put this.

PR #236, opened 2018-05-24
Various fixes

### LINKED ISSUE #215: Include wave name/keyfile content in amplitude hash
So that also waves with different names but equal amplitudes will never have the same hash. Check the code for places where this currently is ignored or only prints a warning and make this a fatal error again (#157).

suhlatwork:
* amplitude file hashes should now be unique (closes #215)
* make isospin-2 states work
* no longer requires an individual wavelist for each mass bin when using binned mass dependencies (probably finishes the last remainders of #50)
* several smaller improvements


stefan-wallner:
Further discussed in #250

PR #237, opened 2018-06-29
Get fit result

stefan-wallner:


partialWaveFit/fitResult.cc
```diff
@@ -387,34 +401,49 @@ fitResult::fill(const fitResult& result,
                 const complexMatrix*            acceptedNormIntegral,    // normalization integral matrix with acceptance
                 const vector<double>*           phaseSpaceIntegral)      // normalization integral over full phase space without acceptance
 {
-	std::vector<std::complex<double>> prodAmpsEmpty;
+	if (fitParCovMatrix == nullptr)
+		fitParCovMatrix = (not result.covMatrixValid()) ? nullptr : &result.fitParCovMatrix();
+	if (fitParCovMatrixIndices == nullptr)
+		fitParCovMatrixIndices = &result.fitParCovIndices();
```
suhlatwork: Why is `fitParCovMatrixIndices` still a pointer in this function?

partialWaveFit/fitResult.cc
```diff
@@ -387,34 +401,49 @@ fitResult::fill(const fitResult& result,
...
 {
-	std::vector<std::complex<double>> prodAmpsEmpty;
+	if (fillCovMatrix)
+		fillCovMatrix = result.covMatrixValid();
+	bool fillNormIntMatrix = false;
+	bool fillAccNormIntMatrix = false;
+	bool fillPhaseSpace = false;
+	if (fillIntegralMatrices){
+		fillNormIntMatrix    = result.normIntegralMatrix().nRows() != 0         or result.normIntegralMatrix().nCols() != 0;
+		fillAccNormIntMatrix = result.acceptedNormIntegralMatrix().nRows() != 0 or result.acceptedNormIntegralMatrix().nCols() != 0;
```
suhlatwork: `or` -> `and`?

partialWaveFit/fitResult.cc
```diff
@@ -243,103 +262,188 @@ fitResult::fill(const unsigned int              nmbEvents,               // numb
...
-	_normIntIndexMap        = result.normIntIndexMap();
-	_phaseSpaceIntegral     = result.phaseSpaceIntegralVector();
-	_converged              = result.converged();
-	_hasHessian             = result.hasHessian();
+	if (fitParCovMatrix == nullptr)
+		fitParCovMatrix = (not result.covMatrixValid()) ? nullptr : &result.fitParCovMatrix();
+	if (fitParCovMatrixIndices == nullptr)
+		fitParCovMatrixIndices = &result.fitParCovIndices();
+	if (normIntegral == nullptr)
+		normIntegral = (result.normIntegralMatrix().nRows() == 0 and result.normIntegralMatrix().nCols() == 0) ? nullptr : &result.normIntegralMatrix();
```
suhlatwork: `and` -> `or`

partialWaveFit/fitResult.cc
```diff
@@ -243,103 +262,188 @@ fitResult::fill(const unsigned int              nmbEvents,               // numb
...
+	if (acceptedNormIntegral == nullptr)
+		acceptedNormIntegral = (result.acceptedNormIntegralMatrix().nRows() == 0 and result.acceptedNormIntegralMatrix().nCols() == 0) ?
```
suhlatwork: `and` -> `or`

partialWaveFit/fitResult.cc
```diff
@@ -243,103 +262,188 @@ fitResult::fill(const unsigned int              nmbEvents,               // numb
...
-	_fitParCovMatrixIndices = result.fitParCovIndices();
-	_normIntegral           = result.normIntegralMatrix();
-	_acceptedNormIntegral   = result.acceptedNormIntegralMatrix();
-	_normIntIndexMap        = result.normIntIndexMap();
-	_phaseSpaceIntegral     = result.phaseSpaceIntegralVector();
-	_converged              = result.converged();
-	_hasHessian             = result.hasHessian();
+	if (fitParCovMatrix == nullptr)
+		fitParCovMatrix = (not result.covMatrixValid()) ? nullptr : &result.fitParCovMatrix();
+	if (fitParCovMatrixIndices == nullptr)
```
suhlatwork: why is this still a pointer here?

partialWaveFit/fitResult.cc
stefan-wallner: I think it is correct his way, because you want to interpret the matrix only as empty if rows==0 and cols ==0.

partialWaveFit/fitResult.cc
stefan-wallner: I think it is correct his way, because you want to interpret the matrix only as empty if rows==0 and cols ==0.
Everything else should raise a warning, as it does this way.

partialWaveFit/fitResult.cc
stefan-wallner: I think it is correct his way, because you want to interpret the matrix only as empty if rows==0 and cols ==0.
Everything else should raise a warning, as it does this way.

partialWaveFit/fitResult.cc
stefan-wallner: It should not! Thank you.

partialWaveFit/fitResult.cc
suhlatwork: No, a matrix with rows==0 and cols!=0 also does not contain any entries.

stefan-wallner: Yes.
However, there is in principle a difference between a result as an empty matrix, which would also be rows==0 and cols!=0, and a result as now matrix.
My proposal as a representation for a result without matrix is rows==0 and cols==0.
Additionally, if the old result has a matrix with rows==0 and cols!=0 there is clearly something wrong.
The question here is whether the fill function should rise a warning in such a situation, which it does at the moment.

suhlatwork: ahh, yes, you are right.

partialWaveFit/fitResult.cc
stefan-wallner: Because this function should give you the ability to copy a fit result and overwrite certain matrices.

suhlatwork: So we will have different types for `fitParCovMatrixIndices` for the various `fill` functions just for a feature (overwriting the mapping of production amplitudes to covariance matrix entries) that is not used anywhere in the current code?

stefan-wallner: This `fill` function, which allows overwriting of matrices, will be used here #238.
Only the `fitParCovMatrixIndices` will not be overwritten, but I think this `fill` function should also have this possibility for completeness.

bgrube:
Merged. Thank you!

PR #238, opened 2018-07-04
Keep matrices only best

stefan-wallner:


pyInterface/module/_fit.py
```diff
@@ -122,6 +126,10 @@ def pwaNloptFit(eventAndAmpFileDict,
 			seeds.append(randVal)
 
 	fitResults = [ ]
+	iBest = -1
+	iBestValid = -1
+	negLogLikeBest = 1e300
+	negLogLikeBestValid = 1e300
```
bgrube: Would it possible to replace 1e300 by "inf" or something similar?

pyInterface/module/_fit.py
```diff
@@ -131,4 +139,71 @@ def pwaNloptFit(eventAndAmpFileDict,
 		                                       saveSpace          = saveSpace,
 		                                       verbose            = verbose)
 		fitResults.append(fitResult)
+		iFitResult = len(fitResults)-1
```
bgrube: It would be nice to also have this also in the pwaFit() function. pwaFit() and pwaNloptFit() actually are very similar and seem to differ only in the call to the fitter. So maybe one could move some of the shared code to service functions.

highLevelInterface/pwaNloptFit.cc
```diff
@@ -25,13 +25,42 @@ double
 rpwaNloptFunc(unsigned int n, const double* x, double* gradient, void* func_data)
 {
 	const pwaLikelihood<complex<double> >* L = (const pwaLikelihood<complex<double> >*)func_data;
-	if(n != L->nmbPars()) {
+	if(n != L->nmbPars() - L->nmbParsFixed()) {
 		printErr << "parameter mismatch between NLopt and pwaLikelihood. Aborting..." << endl;
+		printErr << "Likelihood needs "  << L->nmbPars() << " parameters including " << L->nmbParsFixed() << " fixed parameters. Got " << n << " parameters" << endl;
```
bgrube: Merge printErr message with the line above, i.e. "parameter mismatch between NLopt and pwaLikelihood: likelihood needs "  << ....  "Aborting..."

highLevelInterface/pwaNloptFit.cc
```diff
@@ -25,13 +25,42 @@ double
...
 		throw;
 	}
+	if (L->nmbParsFixed() != 0) {
+		static double* xIncludingFixed = new double[L->nmbPars()];
```
bgrube: I think the code would be easier to read with the following name changes
n -> nmbParsNlopt
x -> parsNlopt
xIncludingFixed -> parsIncludingFixed

highLevelInterface/pwaNloptFit.cc
```diff
@@ -183,21 +215,36 @@ rpwa::hli::pwaNloptFit(const pwaLikelihood<complex<double> >& L,
...
 		timer.Start();
 		nlopt::result result = optimizer.optimize(params, likeli);
-		if(result > 0) {
+		if (result > 0) {
 			converged = true;
 		}
 		timer.Stop();
 
+		// add fixed parameters to params
+		{
```
bgrube: This loop appears 3 times in the code, it would be good to make a function for this.

highLevelInterface/pwaNloptFit.cc
stefan-wallner: Done

highLevelInterface/pwaNloptFit.cc
stefan-wallner: I agree!

highLevelInterface/pwaNloptFit.cc
```diff
@@ -102,12 +131,13 @@ rpwa::hli::pwaNloptFit(const pwaLikelihood<complex<double> >& L,
 		}
 	}
 
-	const unsigned int nmbPar  = L.NDim();
-	const unsigned int nmbEvts = L.nmbEvents();
+	const unsigned int nmbPar       = L.NDim();
```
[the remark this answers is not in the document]
stefan-wallner: Done

highLevelInterface/pwaNloptFit.cc
stefan-wallner: Done. Please check if it is ok this way.

highLevelInterface/pwaNloptFit.cc
```diff
@@ -25,13 +25,42 @@ double
...
+		if (L->nmbParsFixed() != 0) {
+			unsigned int iNlopt = 0;
+			for (unsigned int iRootPWA = 0; iRootPWA < L->nmbPars(); ++iRootPWA) {
+				if (not L->parameter(iRootPWA).fixed()) {
+					gradientNlopt[iNlopt] = gradientIncludingFixed[iRootPWA];
+					++iNlopt;
+				}
+			}
+		}
+
```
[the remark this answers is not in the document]
stefan-wallner: We could do this, but then we would need to copy the gradient from the `gradientIncludingFixed` to `gradient` also in cases without fixed parameters.

bgrube:
Merged. Thank you!

PR #239, opened 2018-08-29
Python version in Boost libraries

suhlatwork:
Since Boost 1.67.0 the Boost.Python library name includes the Python version. Require a more recent Cmake version that supports this.


bgrube:
Thanks, Sebastian, for this fix. It is nice that this makes the hack for the optional Boost components obsolete. The only potential problem I see that this will bring us back to the times, when everyone had to have his private CMake installation. I just had a quick look at the highest available CMake versions:
E18: 3.4
lxplus: 3.7
Ubuntu 18.04: 3.10
MacPorts: 3.12
I guess 3.11 is the lowest version that provides the OPTIONAL_COMPONENTS feature for Boost?

suhlatwork:
3.11 is the lowest version that supports the Python version in the Boost library name.

bgrube:
Included in #241.

PR #246, opened 2019-04-26
Fourier Connected Likelihood

fkaspar:
- Implemented a prior on the fourier transform of the amplitudes.

- Also implemented positivity constraints for the reference waves via NLopt (enabled by default)

### DIFF
--- a/pyInterface/module/pyPartialWaveFit/_startParameterGenerator.py
+++ b/pyInterface/module/pyPartialWaveFit/_startParameterGenerator.py
@@ -30,9 +30,10 @@ class StartParameterGeneratorRpwaUniform(StartParameterGenerator):
 	'''
 	Draw amplitudes uniformly in the range [-sqrt(nevents), sqrt(nevents)]
 	'''
-	def __init__(self, model, seed=None):
+	def __init__(self, model, seed=None, scaleToNbar = True, constrainPosReal = True):
 		StartParameterGenerator.__init__(self, model, seed=seed)
-		self.scaleToNbar = True
+		self.scaleToNbar = scaleToNbar
+		self.constrainPosReal = constrainPosReal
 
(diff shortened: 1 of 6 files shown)


pyInterface/module/pyPartialWaveFit/_startParameterGenerator.py
```diff
@@ -47,6 +48,13 @@ def __call__(self):
 		# this sets the amplitues of real-valued waves (e.g. reference wave) to real
 		paraNegLlhd = [amplVectors, np.array([1]*self.model.nmbDatasets)] + [1]*self.model.parameterMapping.nmbAdditionalParameters
 		paraFitter = self.model.parameterMapping.paraLlhd2Fitter(self.model.parameterMapping.paraNegLlhd2Llhd(paraNegLlhd))
+
+		# change real waves to real AND positive
+		if self.constrainPosReal and hasattr(self.model.parameterMapping, "indicesImagFitterPara"):
+			for i,idxIFP in enumerate(self.model.parameterMapping.indicesImagFitterPara):
+				if idxIFP is None and self.model.parameterMapping.indicesRealFitterPara[i] is not None:
+					paraFitter[self.model.parameterMapping.indicesRealFitterPara[i]] = abs(paraFitter[self.model.parameterMapping.indicesRealFitterPara[i]])
+
```
stefan-wallner: Why to you use the `indicesImagFitterPara` here?
This attribute is only implemented for the `ParameterMappingConnected`.
So setting the `constrainPosReal` for the not-connected likelihood would not work.
If you set the constrain for the fitter, which you do by default, this leads to invalid start parameters for the not-connected likelihood, as far is I can see.

stefan-wallner:
Most of the fixes included in #247

PR #249, opened 2020-05-19
Cleanup and improvement of the build system

bgrube:


cmakeModules/FindPYTHON.cmake
```diff
@@ -165,10 +165,13 @@ if(PYTHONINTERP_FOUND)
 	foreach(_PYTHON_INCLUDE_DIR ${PYTHON_INCLUDE_DIRS})
 		# get library version from patchlevel.h
 		parse_version_from_file("${_PYTHON_INCLUDE_DIR}/patchlevel.h" "#define[ \t]+PY_VERSION[ \t]+" PYTHONLIBS_VERSION)
+		# cut release candidate number, i.e. 4th version number, from PYTHONLIBS_VERSION
+		STRING(REGEX REPLACE "^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)" "\\1.\\2.\\3" PYTHONLIBS_VERSION ${PYTHONLIBS_VERSION})
```
stefan-wallner, line 169: Perhaps `STRING(REGEX REPLACE "^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]*)" "\\1.\\2.\\3" PYTHONLIBS_VERSION ${PYTHONLIBS_VERSION})` would be better, because it catches also the default case and does not ignore it, which results of course in the same string.

cmakeModules/FindROOT.cmake
```diff
@@ -181,35 +175,19 @@ else()
 	string(REGEX REPLACE "^[0-9]+\\.[0-9][0-9]+\\/([0-9][0-9]+).*" "\\1"
 		ROOT_PATCH_VERSION "${ROOT_VERSION}")
 	set(ROOT_VERSION "${ROOT_MAJOR_VERSION}.${ROOT_MINOR_VERSION}.${ROOT_PATCH_VERSION}")
-	# compare version
-	if(ROOT_FIND_VERSION_EXACT)
```
stefan-wallner, line 185: Is there a reason to drop this functionality?

decayAmplitude/ampIntegralMatrix.cc
```diff
@@ -31,7 +31,15 @@
...
 
+#include <boost/version.hpp>
+#if (BOOST_VERSION >= 107200)
+#include <boost/timer/progress_display.hpp>
+using boost::timer::progress_display;
+#else
 #include <boost/progress.hpp>
+using boost::progress_display;
+#endif
+
```
stefan-wallner, line 42: Could we move this code to one header file, as it appears multiple times?

stefan-wallner: Yes, it appears really often ;)

cmakeModules/FindROOT.cmake
bgrube: This is handled by `find_package_handle_standard_args`, i.e. `find_package(ROOT 6.08.00 EXACT ...` should work. Didn't try it though...

decayAmplitude/ampIntegralMatrix.cc
bgrube: Yes, I agree it is ugly now.

cmakeModules/FindPYTHON.cmake
bgrube: Agreed. Will do.

PR #250, opened 2020-05-19
Various fixes

### LINKED ISSUE #215: Include wave name/keyfile content in amplitude hash

stefan-wallner:
* amplitude file hashes should now be unique (closes #215)
* make isospin-2 states work
* no longer requires an individual wavelist for each mass bin when using binned mass dependencies (probably finishes the last remainders of #50)
* several smaller improvements

The original pull request was #236


decayAmplitude/isobarDecayTopology.cc
```diff
@@ -652,7 +657,7 @@ isobarDecayTopology::getIsospinSymmetrization()
 	} // End of the loop over the permutations
 
 	for(unsigned int i = 0; i < symAmplitudes.size(); ++i) {
-		symAmplitudes[i].factor = signum(symAmplitudes[i].factor.real()) / sqrt(symAmplitudes.size());
```
stefan-wallner, line 655: This is clearly something different?

decayAmplitude/isobarDecayTopology.cc
```diff
@@ -567,25 +567,32 @@ isobarDecayTopology::getIsospinSymmetrization()
 		}
 		// Some first checks if the permutation makes sense.
 		bool breaking = false;
-		for(unsigned int j = 0; j < map.size(); ++j) {
```
stefan-wallner, line 570: In the original code, all permutations are checked against the identity, while in the new code, all but the first permutation are checked against the identity (and all previously accepted permutations).
If the first permutation is the identity, which I think it is, this does not matter. However, if the first permutation is a permutation that breaks w.r.t. the identity, this one would be accepted, while the identity would be dropped.

partialWaveFit/pwaLikelihood.cc
```diff
@@ -1363,6 +1394,11 @@ bool
 pwaLikelihood<complexT>::setOnTheFlyBinning(const multibinBoundariesType&       multibinBoundaries,
                                             const vector<const eventMetadata*>& evtMetas)
 {
+	if(not multibinBoundaries.empty()) {
```
stefan-wallner, line 1397: Can we remove the `if` statement?

stefan-wallner (commented):
I went through the changes. Awaiting @bgrube comments.

decayAmplitude/isobarDecayTopology.cc
```diff
@@ -502,159 +506,125 @@ isobarDecayTopology::getIsospinSymmetrization()
...
+		bool discardPermutation = false;
+		for(vector<symTermMap>::const_iterator symAmp = symAmplitudes.begin(); symAmp != symAmplitudes.end(); ++symAmp) {
+
+			if(_debug) {
+				cout << "    ... checking against permutation " << permMapString(symAmp->fsPartPermMap) << endl;
+			}
 
-				if (symAmp->fsPartPermMap[j] == map[j]) {
-					continue; // particle j was not swapped -> go on with check of next particle
+			for(unsigned int i = 0; i < permMap.size(); ++i) {
```
stefan-wallner, line 545: Perhaps ‘i’ -> ‘i_particle’ would be better, because the body of the loop is rather long.

stefan-wallner (commented):
The code is much clearer now. Thank you.

stefan-wallner:
Looks good to me.
Anything missing from your side, @bgrube ?

bgrube:
From my side this is ready for merging.

stefan-wallner:
Thank you.

